PR #2, opened 2020-05-07
Patient CSV Extractor

Dtphelan1:
# Summary 
Creates a CSV format for patients and an extractor that uses this format. When this PR is merged, we'll update the ICARECSVClient to use this CSVPatientExtractor. 

## New Functionality
Exposes a new CSVExtractor that can parse CSV's containing patient information. 

## Code Changes
- Updates the EJS generateMcodeResources function so that ID's on the incoming data take precedence over the auto-generated ones
- Same with the RenderTemplate function 
- Updated the intermediary representation of patient data that's fed into our `Patient.R4.ejs` template
- Within the template tests, updated the `patient.test.js`  file to align with the new intermediary format

## Testing
- Review the CSV format itself.
- Ensure tests pass
- Use the fixture within the extractor tests to confirm that the output FHIR is valid.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,4 @@
+# Summary
+## New behavior
+## Code changes
+# Testing guidance
--- a/docs/patient.csv
+++ b/docs/patient.csv
@@ -1,0 +1,3 @@
+mrn,patientId,family,given,gender
+pat-mrn-1,pat-id-1,last,first,female
+pat-mrn-2,pat-id-2,final,preceding,male
--- a/src/extractors/CSVPatientExtractor.js
+++ b/src/extractors/CSVPatientExtractor.js
@@ -25,7 +25,7 @@ class CSVPatientExtractor {
   async getPatientData(mrn) {
     logger.info('Getting patient data');
     const data = await this.csvModule.get('mrn', mrn);
-    // Should only need one patient
+    // Should only be one patient with this mrn; get that pat from our arr
     return data[0];
   }
 
--- a/src/extractors/CSVPatientExtractor.js
+++ b/src/extractors/CSVPatientExtractor.js
@@ -3,6 +3,7 @@ const { generateMcodeResources } = require('../helpers/ejsUtils');
 const logger = require('../helpers/logger');
 
 function joinAndReformatData(patientData) {
+  logger.info('Reformatting patient data from CSV into template format');
   // No join needed, just a reformatting
   const {
     mrn, patientId, family, given, gender,
--- a/src/extractors/CSVPatientExtractor.js
+++ b/src/extractors/CSVPatientExtractor.js
@@ -1,5 +1,6 @@
 const { CSVModule } = require('../modules');
 const { generateMcodeResources } = require('../helpers/ejsUtils');
+const { Extractor } = require('./Extractor');
 const logger = require('../helpers/logger');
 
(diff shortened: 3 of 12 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,4 @@
+# Summary
+## New behavior
+## Code changes
+# Testing guidance
```
mgramigna, line 4: Look at us :)

Dtphelan1: They said it couldn't be done.

src/extractors/CSVPatientExtractor.js
```diff
@@ -0,0 +1,47 @@
...
+  return {
+    id: patientId,
+    mrn,
+    family,
+    given,
+    gender,
+  };
+}
+
+class CSVPatientExtractor {
```
mgramigna, line 21: [MG1] Should this extend extractor?

src/helpers/ejsUtils.js
```diff
@@ -34,7 +35,7 @@ function generateMcodeResources(mcodeProfileID, data) {
     resourceType: 'Bundle',
     type: 'collection',
     entry: (_.isArray(data) ? data : [data]).map((d) => ({
-      fullUrl: `urn:uuid:${generateResourceId(d)}`,
+      fullUrl: `urn:uuid:${d.id ? d.id : generateResourceId(d)}`,
```
mgramigna, line 38: [NITPICK] Could just use `d.id || generateResourceId` here

src/extractors/CSVPatientExtractor.js
Dtphelan1: Ew that's a great idea.

src/helpers/ejsUtils.js
Dtphelan1: Addressed

src/extractors/CSVPatientExtractor.js
Dtphelan1: Addressed

docs/patient.csv
```diff
@@ -0,0 +1,3 @@
+mrn,patientId,family,given,gender
```
jafeltra, line 1: I might be getting lost in all the ids, but should this align with the others you changed in #3 and be `subjectId` rather than `patientId`? There's a good chance I'm losing track of each id, so maybe this is fine as is.

Dtphelan1: Great catch, unclear on the answer. The weridness here is that the ID isn't for the subject of a condition or an observation, it's for the patient proper. I'm open to using `subjectId` ubiquitously, but I wouldn't be shocked if there's some subtle different in FHIR-land that we'd be missing. Then again, any alignment between our CSV format and FHIR is a nice to have; all that we need to be prioritizing is internal consistency.

jafeltra: Yeah I think I was worried I was missing a subtle difference, which I think I was. Since this is the actual id of the resource and not a reference to a different resource, I think maybe making it subjectId will actually confuse things because it will seem like both places should serve the same purpose. So I think maybe keeping patientId is best.

Dtphelan1:
Rebased

Dtphelan1:
Rebased again after v2 release

jafeltra:
All looks good after rebases and there's two approvals so I'll merge

PR #4, opened 2020-05-12
Condition CSV Extractor

danlee1025:
# Summary
Added CSV format for condition information and extractor to retrieve data.
## New behavior
New CSV extractor to extract condition data
## Code changes
- Added EJS template and test for Condition resource
    - `category` and `verificationStatus` were required fields. I gave them values of `encountered-diagnosis` and `confirmed` in the template. Did we want different values here? Did we want to include this in the CSV format?
- Added `CSVConditionExtractor`
# Testing guidance
- Review CSV format
    - I put `codeSystem` and `code` as separate values for easier parsing. Did we want to combine these into one value that we would parse (e.g. value would be `system: code`)?
    - Did we want to only allow codes from one system in which case we would only have a `code` value in the csv format?
- Ensure tests pass

### DIFF
--- a/src/templates/Condition.ejs
+++ b/src/templates/Condition.ejs
@@ -1,0 +1,52 @@
+<%# Condition Template %>
+<%# Based on https://mcodeinitiative.github.io/StructureDefinition-obf-Condition.html %>
+<%# Official url: http://hl7.org/fhir/us/mcode/StructureDefinition/obf-Condition %>
+<% /* %>
+{
+  id: String,
+  subject: {
+    id: String,
+    name: String,
+  },
+  code: {
+    system: String,
+    code: String,
+    display: String,
+  },
+}
+<% */ %>
+{
+  "resourceType": "Condition",
+  "id": "<%- id %>",
+  "subject" : {
+    "reference": "Patient/<%- subject.id %>"<% if (subject.name) { %>,
+    "display": "<%- subject.name %>"
+    <% } %>
+  },
+  "code": {
+    "coding": [
+      {
+        "system": "<%- code.system %>",
+        "code": "<%- code.code %>"<% if (code.display) { %>,
+        "display": "<%- code.display %>"
+        <% } %>
+      }
+    ]
+  },
+  "category": {
+    "coding": [
+      {
+        "system": "http://hl7.org/fhir/us/core/ValueSet/us-core-condition-category",
+        "code": "encounter-diagnosis"
+      }
+    ]
+  },
+  "verificationStatus": {
+    "coding": [
+      {
+        "system": "http://hl7.org/fhir/ValueSet/condition-ver-status",
+        "code": "confirmed"
+      }
+    ]
+  }
+}
--- a/src/templates/Condition.ejs
+++ b/src/templates/Condition.ejs
@@ -37,7 +37,7 @@
     "coding": [
       {
         "system": "http://hl7.org/fhir/us/core/ValueSet/us-core-condition-category",
-        "code": "encounter-diagnosis"
+        "code": "problem-list-item"
       }
     ]
   },
--- a/src/templates/Condition.ejs
+++ b/src/templates/Condition.ejs
@@ -33,18 +33,20 @@
       }
     ]
   },
-  "category": {
-    "coding": [
-      {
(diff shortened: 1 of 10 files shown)


Dtphelan1:
DP1 - We should change the `encountered-diagnosis` to `problem-list-item` AFAIK

src/templates/Condition.ejs
```diff
@@ -0,0 +1,52 @@
...
+        "code": "<%- code.code %>"<% if (code.display) { %>,
+        "display": "<%- code.display %>"
+        <% } %>
+      }
+    ]
+  },
+  "category": {
+    "coding": [
+      {
+        "system": "http://hl7.org/fhir/us/core/ValueSet/us-core-condition-category",
```
jafeltra, line 39: I think the system should actually be: http://terminology.hl7.org/CodeSystem/condition-category

Looking at the link you have here, it lists that link as the `system` (plus it validated with that link).

src/templates/Condition.ejs
```diff
@@ -0,0 +1,52 @@
...
+        "code": "problem-list-item"
+      }
+    ]
+  },
+  "verificationStatus": {
+    "coding": [
+      {
+        "system": "http://hl7.org/fhir/ValueSet/condition-ver-status",
```
jafeltra, line 47: I seem to be getting validation errors with this system url also, but I haven't figured out what the right url is.

jafeltra: Update: I think it's supposed to be this link http://terminology.hl7.org/CodeSystem/condition-ver-status (which I just found by looking at a US Core Condition example)

src/templates/Condition.ejs
Dtphelan1: Could we actually use the URL the above redirects to instead? https://terminology.hl7.org/1.0.0/CodeSystem-condition-category.html

src/templates/Condition.ejs
Dtphelan1: Could we actually use the URL the above readirects to instead? 
https://terminology.hl7.org/1.0.0/CodeSystem-condition-ver-status.html

src/templates/Condition.ejs
```diff
@@ -0,0 +1,52 @@
...
+    "coding": [
+      {
+        "system": "<%- code.system %>",
+        "code": "<%- code.code %>"<% if (code.display) { %>,
+        "display": "<%- code.display %>"
+        <% } %>
+      }
+    ]
+  },
+  "category": {
```
jafeltra, line 36: One last validation error - I think `category` should be an array, so this object should just get wrapped into an array.

jafeltra:
Besides those few validation errors, this looks good to me!

src/templates/Condition.ejs
jafeltra: Dan pointed out that the obf-condition profile makes this property `1..1`. I'm not 100% sure, but I think even if the profile constrains the cardinality, the resource should still have the original array structure. I'm going to try to confirm this though.

Dtphelan1:
Was talking with Jules about these validation errors, but you've already addressed most of them. All told this looks pretty great 😄

test/templates/fixtures/condition-resource.json
```diff
@@ -0,0 +1,31 @@
...
+      {
+        "system": "http://hl7.org/fhir/us/core/ValueSet/us-core-condition-category",
+        "code": "problem-list-item"
+      }
+    ]
+  },
+  "verificationStatus": {
+    "coding": [
+      {
+        "system": "http://hl7.org/fhir/ValueSet/condition-ver-status",
```
jafeltra, line 26: Do you think the same changes should get put into this resource also?

src/templates/Condition.ejs
jafeltra: Just to document it here, I found this documentation https://www.hl7.org/fhir/profiling.html#cardinality, which says

>Note that though a profile can constrain an element from x..* to x..1, this doesn’t make any difference to the representation in the JSON format - the element will still be represented in an array

test/templates/fixtures/condition-resource.json
danlee1025: good catch! I updated the test to actually check for it

PR #5, opened 2020-05-15
Separate Clinical Trial Info extractors

jafeltra:
# Summary
Separates a CSV-only and Service-CSV Clinical Trial Info extractors

## New behavior
*UPDATE: This has been moved to the Epic-MEF.* `ServiceClinicalTrialInfoExtractor` is the same as master's `CSVClinicalTrailExtractor`. When we add context into all the Epic extractors, we will add it into this extractor. We'll first try to get patient/patientId from context, and if it is not available, we'll use the BaseFHIRModule to get patient. Since I didn't start any of the context tasks, it just uses BaseFHIRModule as before. **Edit:** This now lives in the Epic-MEF and there's a [PR here](https://gitlab.mitre.org/mcode/epic-mcode-extraction-framework/-/merge_requests/64).

`CSVClinicalTrialExtractor` uses no services at all. When/If we add context to non-Epic extractors, we will add it into this extractor. We'll first try to get patient/patientId from context, and if it is not available, we'll use the MRN as a default id. Since I didn't start any of the context tasks, I just wrote a function that does nothing right now and then defaults to MRN.

## Code changes
Renamed `CSVClinicalTrailInfoExtractor` to `ServiceClinicalTrialInfoExtractor`, created new `CSVClinicalTrialInfoExtractor` without dependency on BaseFHIRModule. Added corresponding updates to tests.

# Testing guidance
Check the code/tests. There is a related [PR in icare-extraction-client](https://gitlab.mitre.org/mcode/icare-extraction-client/-/merge_requests/17) if you want to try running the CSV client to see it get past the Clinical Trial Extraction successfully.

One note - I didn't start any of the context pieces in this PR because it seemed to require a number of the subtasks we have in the context task, so I tried to lay out where some of the pieces will go when we do that task.


Dtphelan1:
DP1 - Talked a bit offline, but posting here to centralize the next steps: I think the ServiceClinInfoExt should live in the other EHR-specific repo, and should be renamed accordingly. The way the Extractor is designed is based on the constraints of that EHR-specific env and the information we have about what WebServices are and are not available in that env. Even though none of the modules it uses are proprietary to that env, the choices we've  made about what modules we use and in what order are informed by the EHR's constraints.

test/extractors/CSVClinicalTrialInformationExtractor.test.js
```diff
@@ -45,10 +21,11 @@ csvModuleSpy
 
 test('clinical trial information extractor', async () => {
   const data = await csvClinicalTrialInformationExtractor.get({ mrn: MOCK_PATIENT_MRN });
+  const exampleBundleWithMRN = { ...exampleClinicalTrialInformationBundle };
```
Dtphelan1, line 24: DP2? - Not 100% sure why this line is absolutely needed. I can see that this is cloning the bundle... did we think we were mutating it and wanted to ensure we didn't mutate the original? It doesn't look like the code does this AFAICanTell

jafeltra: Oops yeah I was manipulating it, but then made a different fixture and forgot to take this out. I can remove it.

test/extractors/ServiceClinicalTrialInformationExtractor.test.js
```diff
@@ -0,0 +1,54 @@
...
+const exampleClinicalTrialInformationBundle = require('./fixtures/service-clinical-trial-information-bundle.json');
+
+// Constants for mock tests
+const MOCK_URL = 'http://localhost';
+const MOCK_CSV_PATH = path.join(__dirname, 'fixtures/example.csv'); // need a valid path/csv here to avoid parse error
+const MOCK_REQUEST_HEADERS = {
+  Accept: 'application/json',
+};
+
+const MOCK_PATIENT_MRN = 'EXAMPLE-MRN';
```
Dtphelan1, line 15: DP3 - I don't think that this MRN matches the MRN in the `examplePatient` fixture. Which I kind of find confusing... I generally find our extractor tests pretty damn confusing. Do you think that in the extractor test we could check that the expected properties from the various module responses are used where we expect them to be used? For example: In the service task could we check that Id on the examplePatient matches the ID on the resulting data, and that the data provided by the ClinicalTrialModule is where we expect it in the resulting data? I think that's probably what we should be doing instead of checking against a static fixture, since we're testing the Extractors ability to combine different data sources. 

Also if you feel like this change in testing approach is radically outside the original scope of your task, we can discuss creating alternative tasks. My vote would be for us to update the other tests in a separate task if the trial run of modifying this test here goes well.

Also open to you pushing back more dramatically. This is me thinking outloud a bit.

jafeltra:
I think DP1 makes sense. I didn't really think about how the combination/steps we're doing are what make it Epic specific, and was just thinking about how the modules are not Epic specific. If somehow we discover some day this is a typical process, we can always move it back into the base repo.

I'm also going to fix DP2 now and start looking into DP3.

jafeltra:
DP1 and DP2 are fixed here.

There's also now [another PR for the Epic MEF](https://gitlab.mitre.org/mcode/epic-mcode-extraction-framework/-/merge_requests/64), which has the files moved as part of DP1.

The icare-extraction-client PR is also updated to reflect changes from DP1.

jafeltra:
3a0958a should address DP3 test-related comments.

Dtphelan1:
There's one more change incoming here, but currently this looks good.

PR #8, opened 2020-06-04
Context Aware Extractors

danlee1025:
# Summary
Extractors now parse context bundle for resources

## New behavior
- Added helper function in `fhirUtils` to use `fhirpath` to search for resources in bundle
- Added logic to parse context bundle in `BaseFHIRExtractor` and `CSVClinicalTrialInformationExtractor`
- Added tests to make sure resources are being retrieved from context instead of from fhir module.
# Testing guidance
- Ensure tests pass
- Ensure quality of the code
- Run the CSV Client
   - Make sure to update the `package.json` to point to the `context-aware` branch.
   - In the log statements when the `CSVClinicalTrialInformationExtractor` is running, it should log that patient resource is found in context.
   - The resulting bundle generated by the client should use the patient id for the research subject instead of the mrn.

### DIFF
--- a/src/extractors/CSVClinicalTrialInformationExtractor.js
+++ b/src/extractors/CSVClinicalTrialInformationExtractor.js
@@ -29,7 +29,13 @@ function joinClinicalTrialData(patientId, clinicalTrialData) {
 
 function getPatientId(context) {
   const patientInContext = getBundleResourcesByType(context, 'Patient', {}, true);
-  return patientInContext ? patientInContext.id : undefined;
+  if (patientInContext) {
+    logger.info('Patient resource found in context.');
+    return patientInContext.id;
+  }
+
+  logger.info('No patient resource found in context.');
+  return undefined;
 }
 
 class CSVClinicalTrialInformationExtractor extends Extractor {
--- a/src/extractors/CSVClinicalTrialInformationExtractor.js
+++ b/src/extractors/CSVClinicalTrialInformationExtractor.js
@@ -1,7 +1,7 @@
 const path = require('path');
 const { Extractor } = require('./Extractor');
 const { CSVModule } = require('../modules');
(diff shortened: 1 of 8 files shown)


src/index.js
```diff
@@ -47,6 +53,7 @@ module.exports = {
   firstResourceInBundle,
   getDiseaseStatusCode,
   generateMcodeResources,
+  getBundleResourcesByType,
   getICD10Code,
   getPatientName,
   isBundleEmpty,
```
Dtphelan1, line 59: DP1 - Could we put the getBundleResources before getDiseaseStatus to maintain alpha-sort?

src/extractors/CSVClinicalTrialInformationExtractor.js
```diff
@@ -27,12 +27,15 @@ function joinClinicalTrialData(patientId, clinicalTrialData) {
...
-  // If patient, get id off of that, return id;
-  // If no patient, return;
+function getPatientId(context) {
+  const patientInContext = getBundleResourcesByType(context, 'Patient', {}, true);
+  if (patientInContext) {
+    logger.info('Patient resource found in context.');
+    return patientInContext.id;
+  }
+
+  logger.info('No patient resource found in context.');
```
Dtphelan1, line 37: DP2 - do we want to move this fn into the base Extractor class? Is that too much? This feels identical to the functionality being done in BaseFhirExtractor, only with additional logs, so I'm wondering if we can simplify. Open question honestly.

Dtphelan1: Ehhhh on second glance, the common need is that we get a patient resource off of context. I'm fine leaving this as is. We have most of the funcitnoality stubbed out in the form of the getBundleResourcesByType anyway. Disregard.

src/index.js
danlee1025: Addressed.

PR #9, opened 2020-06-05
Update FHIR resource parameters based on app orchard requirements

jafeltra:
# Summary
This PR updates parameters for FHIR resource calls, adding additional support to extractors that require additional parameters.

## New behavior

Adds support for supplying parameters in additional FHIR resource extractors. Also, a lot of the medication information for patients in App Orchard was stored in MedicationRequests, so I added an extractor for those as well.

## Code changes

Updates to enable additional parameters and tests for those changes.

# Testing guidance

Run the tests and look through the code.

When using this PR, along with the two in the other repos, you should be able to successfully retrieve FHIR resources. Clinical Notes and SDEs still don't work, so I've been commenting those calls out when testing extraction locally.


src/extractors/FHIRMedicationRequestExtractor.js
```diff
@@ -0,0 +1,25 @@
...
+  async parametrizeArgsForFHIRModule({ mrn, context }) {
+    const paramsWithID = await super.parametrizeArgsForFHIRModule({ mrn, context });
+    // Only add status to parameters if it has been specified
+    return {
+      ...paramsWithID,
+      ...(this.status && { status: this.status }),
+    };
+  }
+}
+
```
Dtphelan1, line 22: DP1'ish - This implementation works, but I was wondering if the API returns all statuses if the value for `status` is undefined. If so, then couldn't we just make the BASE_STATUSES value here undefined, and avoid the weird spread operator on an expression that returns undefined unless there's a status in which case it returns {status: this.status}. Let me know if that makes sense. I think we'd want to test the API endpoint to be sure.

test/extractors/FHIRAllergyIntoleranceExtractor.test.js
```diff
@@ -1,11 +1,38 @@
...
+  test('parametrizeArgsForFHIRModule should add category to param values', async () => {
+    // Create spy
+    const { baseFHIRModule } = extractor;
+    const baseFHIRModuleSearchSpy = jest.spyOn(baseFHIRModule, 'search');
+    baseFHIRModuleSearchSpy.mockReturnValue(examplePatientBundle);
+
+    const params = await extractor.parametrizeArgsForFHIRModule({ mrn: MOCK_MRN });
+    expect(params).toHaveProperty('clinical-status');
+    expect(params['clinical-status']).toEqual(baseClinicalStatus);
+  });
```
Dtphelan1, line 37: Only other thing DP2ish is definitely not on you but would be a nice improvement: instead of having Constructor/fnName at the beginning of every test, could we just wrap similar tests in another `describe`, which just describes which function. So for example this block would start like this: 
```
describe('FHIRAllergyIntoleranceExtractor', () => {
  describe('Constructor', () => 
    test('sets resourceType as AllergyIntolerance', () => {
      expect(extractor.resourceType).toEqual('AllergyIntolerance');
    });
    ... 
  });
```

Dtphelan1:
Other than DP1/2 this lgtm

src/extractors/FHIRMedicationRequestExtractor.js
jafeltra: I did test with the API and if we pass in an undefined status, axios will just put in an empty `status=` parameter into the API call. If we were to have a `status=''` or `status=undefined`, the returned resource will say that an invalid status was supplied. Even though we don't get to that case by always adding `status` here, I thought it was a little nicer to just not include status at all in the parameters if it wasn't defined. I can have it just always add an undefined status if you'd prefer.

test/extractors/FHIRAllergyIntoleranceExtractor.test.js
jafeltra: Sure. I was trying to just follow the existing pattern, but I can update it. Do you want me to update all the test files that follow this pattern?

src/extractors/FHIRMedicationRequestExtractor.js
Dtphelan1: Nahh let's stick with it. Appreciate the test 😄

test/extractors/FHIRAllergyIntoleranceExtractor.test.js
Dtphelan1: All files you touch would be _nice_, only if it's not too much work. SEriously, feel free to push back.

Dtphelan1:
DP1 is sufficiently resolved 👍

test/extractors/FHIRAllergyIntoleranceExtractor.test.js
jafeltra: DP2 should be addressed in eecce05. I didn't wrap the tests in some files where it seemed the extra describe was overkill (one or two tests for the most part), but let me know what you think.

Dtphelan1:
Damn those looks _good_. Thanks Jules. DP2 addressed and lgtm

Dtphelan1:
Needs a rebase.

PR #11, opened 2020-06-08
Updated profiles

Dtphelan1:
Straightforward - updated profiles to reference correct URL's. Bundles created using these resources should now validate on the `icare-ext-client` repo. 

Testing: 
- Verify manually that these are the canonical URL's listed on both IG's, found here:
    - http://standardhealthrecord.org/guides/icare/StructureDefinition-icare-care-plan-with-review.html 
    - http://hl7.org/fhir/us/mcode/StructureDefinition-mcode-cancer-disease-status.html


jafeltra:
This looks good! I have one question, but it might be out of scope for this PR. Should we update the two value sets for primary and secondary disease status? I'm guessing the content hasn't changed, but the files we have have the old standardhealthrecord URLs.

Dtphelan1:
It's a great question, I'd call it out of scope and we should take about it with Rob. Thoughts?

PR #13, opened 2020-06-10
Changed templates to use generic reference ids

danlee1025:
# Summary
Changed templates to use generic reference ids
## New behavior/Code changes
- Updated templates and test fixtures to use `urn:uuid` as prefix for reference ids instead of `Patient/` or `Condition/`.
# Testing guidance
Ensure tests are good and other repos work fine with these changes.

### DIFF
--- a/src/templates/CancerDiseaseStatus.ejs
+++ b/src/templates/CancerDiseaseStatus.ejs
@@ -50,13 +50,13 @@
   },
   "focus" : [
     {
-      "reference": "Condition/<%- condition.id %>"<% if (condition.name) { %>,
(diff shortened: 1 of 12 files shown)


danlee1025:
@jafeltra I updated the csv file so that the condition ids match up. Can you double-check to make sure that warning goes away? I don't think this really matters since it's possible the reference isn't actually in the bundle.

@Dtphelan1 Do you think we still want this? Looking at the thread with Jules and Rob, it seems we might reference non-FHIR resources so I'm not sure we want to provide a type.

Dtphelan1:
I'm good leaving the type out for now!

PR #16, opened 2020-06-18
Cds evidence csv

Dtphelan1:
# Summary
Adds evidence as a new required parameter of the CDS template and augments the CSV-CDS extractor to parse the evidence appropriately  

## New behavior
CSV files containing CDS data must now have an evidence column, and a value that contains a pipe-delimited list of codes associated with evidences coming from the following value set: http://hl7.org/fhir/us/mcode/ValueSet-mcode-cancer-disease-status-evidence-type-vs.html

## Code changes
Update tests to supply this newly required piece of data; augment the EJS template to handle the iteration over a list of evidences and add them to the rendered template accordingly; adding a evidence-type code lookup akin to the disease-status code lookup

# Testing guidance
Ensure test pass, linting passes, and that the ICARE-extraction-client's sibling branch works as expected. 

Probably want two sets of eyes on this since it's a pretty critical new piece of functionality

### DIFF
--- a/src/extractors/CSVCancerDiseaseStatusExtractor.js
+++ b/src/extractors/CSVCancerDiseaseStatusExtractor.js
@@ -54,7 +54,8 @@ class CSVCancerDiseaseStatusExtractor {
     const packagedDiseaseStatusData = joinAndReformatData(diseaseStatusData);
 
     // 3. Generate FHIR Resources
-    return generateMcodeResources('CancerDiseaseStatus', packagedDiseaseStatusData);
+    const resources = generateMcodeResources('CancerDiseaseStatus', packagedDiseaseStatusData);
+    return resources;
   }
 }
 
--- a/src/extractors/CSVCancerDiseaseStatusExtractor.js
+++ b/src/extractors/CSVCancerDiseaseStatusExtractor.js
@@ -1,6 +1,6 @@
 const path = require('path');
 const { CSVModule } = require('../modules');
-const { getDiseaseStatusDisplay } = require('../helpers/diseaseStatusUtils');
+const { getDiseaseStatusDisplay, getDiseaseStatusEvidenceDisplay } = require('../helpers/diseaseStatusUtils');
 const { generateMcodeResources } = require('../helpers/ejsUtils');
 const { formatDateTime } = require('../helpers/dateUtils');
 const logger = require('../helpers/logger');
@@ -13,6 +13,7 @@ function joinAndReformatData(arrOfDiseaseStatusData) {
       throw new Error('DiseaseStatusData missing an expected property: mrn, conditionId, diseaseStatus and dateOfObservation are required.');
     }
   });
+  const evidenceDelimiter = '|';
   return arrOfDiseaseStatusData.map((record) => ({
     // We have no note to base our ObservationStatus off of; default to 'final'
     status: 'final',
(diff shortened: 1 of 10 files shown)


test/templates/cancerDiseaseStatus.test.js
```diff
@@ -19,6 +19,7 @@ const VALID_DATA = {
     name: 'Walking Corpse Syndrome',
   },
   effectiveDateTime: '1994-12-09T09:07:00Z',
+  evidence: null,
```
danlee1025, line 22: Should we add evidence to the test and fixture to make sure it gets generated correctly by the template?

src/extractors/CSVCancerDiseaseStatusExtractor.js
```diff
@@ -49,7 +54,8 @@ class CSVCancerDiseaseStatusExtractor {
     const packagedDiseaseStatusData = joinAndReformatData(diseaseStatusData);
 
     // 3. Generate FHIR Resources
-    return generateMcodeResources('CancerDiseaseStatus', packagedDiseaseStatusData);
```
danlee1025, line 52: Should we change this back to just a single return statement or should we leave it to make it easier to log in the future? I'm assuming you set the const and then return to log the result during testing.

src/helpers/ejsUtils.js
```diff
@@ -25,7 +25,8 @@ function generateResourceId(data) {
 
 function renderTemplate(template, data) {
   // Ensure that spread operator on data is last, so any data.id takes precedence
-  return JSON.parse(ejs.render(template, { id: generateResourceId(data), ...data }));
```
danlee1025, line 28: Same comment as above for this one

danlee1025:
I know we didn't have tests in diseaseStatusUtils but should we add some to make sure the lookups are doing what we expect?

Overall this looks very good, just small comments.

src/extractors/CSVCancerDiseaseStatusExtractor.js
Dtphelan1: I'm thinking of leaving it for future debugging. I've gone back and forth on this one so many times, so I think leaving it would be valuable. Let me know if you disagree!

src/helpers/ejsUtils.js
Dtphelan1: See above :)

jafeltra:
This looks good! I didn't see anything beyond what Dan pointed out!

test/templates/cancerDiseaseStatus.test.js
Dtphelan1: Addressed DL1 😄

Dtphelan1:
Should have addressed all of DL's comments! Let me know if anything else stands out 👍

src/extractors/CSVCancerDiseaseStatusExtractor.js
jafeltra: I don't think I have a strong preference on this, so I think I'd rather wait to see what Dan thinks.

Dtphelan1:
Created tests for diseaseStatus, and discussed with JA re: my approach. Happy to elaborate if needed.

src/extractors/CSVCancerDiseaseStatusExtractor.js
danlee1025: Sounds good to me!

PR #17, opened 2020-06-18
Support filtering by fromDate and toDate in CSV module

jafeltra:
# Summary
Adds support for filtering out CSV data based on a provided `fromDate` and `toDate`.

## New behavior
You can now run the client with a `--fromDate` and `--toDate` specified for CSV extraction. The fromDate (or one from the log file if not provided) and the toDate (if provided) are then used to filter out data in the CSV module `get`. If all data is filtered out, we return an empty collection bundle.

## Code changes
Added filtering to CSV Module `get` based on a `fromDate` and `toDate`, if provided and valid dates. Return an empty bundle if no patient information is returned from extractor's `get`. Added tests for both of those changes. I also updated the joinAndReformat test in the CancerDiseaseStatus tests to more closely mirror the ones in TreatmentPlanChange and ClinicalTrialInfo since it was functionally the same, but I can revert that if we prefer the old way.

# Testing guidance
You should be able to run the client and specify a `--fromDate` and `--toDate` and only have resources for data with a `dateRecorded` after the specified `--fromDate` and before the specified `--toDate`. Also, if you have successfully run the CSV client, it will pull the latest date to use as `fromDate` if it is not specified.

### DIFF
--- a/src/modules/CSVModule.js
+++ b/src/modules/CSVModule.js
@@ -18,7 +18,7 @@ class CSVModule {
     // If fromDate and toDate is provided, filter out all results that fall outside that timespan
     if (fromDate && moment(fromDate).isValid()) result = result.filter((r) => !(r.dateRecorded && moment(fromDate).isAfter(r.dateRecorded)));
     if (toDate && moment(toDate).isValid()) result = result.filter((r) => !(r.dateRecorded && moment(toDate).isBefore(r.dateRecorded)));
-    if (result.length === 0) logger.warn('No data for patient after specified fromDate');
+    if (result.length === 0) logger.warn('No data for patient within specified time range');
     return result;
   }
 }
--- a/src/modules/CSVModule.js
+++ b/src/modules/CSVModule.js
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const moment = require('moment');
 const parse = require('csv-parse/lib/sync');
 const logger = require('../helpers/logger');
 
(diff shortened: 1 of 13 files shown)


Dtphelan1:
Outside your immediate scope, but it would be nice if we introduced the describe motif into the get tests for the modules. Up to you if you wanna take that on. I think it's not so bad, but understand if you wanna keep a strong dividing line

jafeltra:
Haha I think I already smudged away that dividing line, so I'll add describes!

Dtphelan1:
DP1 - You should update the CSV example files in our `docs` file for our `CancerDiseaseStatus` `TreatmentPlanChange `

jafeltra:
DP1 is resolved!

Dtphelan1:
DP2 - We may benefit from a FHIRUtils helper function that is responsible for creating empty bundles and using it throughout our empty-response checkers within CSVExtractors.

src/modules/CSVModule.js
```diff
@@ -7,12 +8,17 @@ class CSVModule {
...
   }
 
-  async get(key, value) {
+  async get(key, value, fromDate, toDate) {
     logger.info(`Get csvModule info by key '${key}'`);
     // return all rows if key and value aren't provided
     if (!key && !value) return this.data;
-    const result = this.data.filter((d) => d[key] === value);
+    let result = this.data.filter((d) => d[key] === value);
     if (result.length === 0) throw new ReferenceError(`CSV Record with provided key '${key}' and value was not found`);
```
Dtphelan1, line 16: DP-_NOT_-your-problem-now: We should consider whether or not we want an error thrown here, or if a logger error would be sufficient.

jafeltra: That's a good point. Now that we better handle returning an empty bundle, maybe a `logger.error` is sufficient or maybe it's something we should really throw an error for. I'll make a note to discuss at backlog grooming.

Dtphelan1:
I think this looks good!

src/modules/CSVModule.js
```diff
@@ -7,12 +8,17 @@ class CSVModule {
...
+
+    // If fromDate and toDate is provided, filter out all results that fall outside that timespan
+    if (fromDate && moment(fromDate).isValid()) result = result.filter((r) => !(r.dateRecorded && moment(fromDate).isAfter(r.dateRecorded)));
+    if (toDate && moment(toDate).isValid()) result = result.filter((r) => !(r.dateRecorded && moment(toDate).isBefore(r.dateRecorded)));
+    if (result.length === 0) logger.warn('No data for patient after specified fromDate');
```
kjmahalingam, line 21: DM1: This warning seems misleading for the case where there is data after the specified fromDate, but not *before the specified toDate*. Can we update this message to be more generic? Or if it's easy, have two separate messages for both cases, but I wouldn't spend too much effort on that.

jafeltra: Oops yeah I remembered to update the comment when I added `toDate` but forgot to update the logger message. Whooops.

jafeltra:
DM1 should be fixed.

PR #21, opened 2020-07-24
Condition extractor

jafeltra:
# Summary

Adds a new helper function to get full entries of bundles.

## New behavior
None

## Code changes
New function

# Testing guidance
This is really to support the PRs in the other two repos. Using this along with those branches should successfully bundle Conditions.


src/helpers/fhirUtils.js
```diff
@@ -56,11 +56,26 @@ const getBundleResourcesByType = (bundle, type, context = {}, first) => {
   return first ? null : [];
 };
 
+const getBundleEntriesByResourceType = (bundle, type, context = {}, first) => {
```
Dtphelan1, line 59: DP1 : super small, but could we, for the sake of prettiness, make false the default value for first?

Dtphelan1, line 59: DP1 : super small, but could we, for the sake of prettiness, make false the default value for first?

```suggestion
const getBundleEntriesByResourceType = (bundle, type, context = {}, first=false) => {
```

Dtphelan1:
Bar that small nitpick, this looks great! Nice work, and a super useful helper 👍

src/helpers/fhirUtils.js
kjmahalingam: I'm not really reviewing this PR so feel free to ignore me trespassing, but if we do that, can we make that change for `getBundleResourcesByType` as well?

Dtphelan1: yes please.

jafeltra: Sure! I think I was going off of the `getBundleResourcesByType` function, but I can update both!

PR #22, opened 2020-07-29
Update to debug statements

jafeltra:
# Summary
Updates all `logger.info` statements to be `logger.debug` statements. 

## New behavior
No new code behavior, but limits the number of statements that get logged when running the CLI using extractors in this repo.

## Code changes
Uses `logger.debug` instead of `logger.info`. I also tried to align the logging statements across extractors, so I added a few debug statements in extractors to match those in the other extractors.

One small side effect is that the `log.warn` statements stand out a little more, but I think they all seem valid as warnings, so I left them.

# Testing guidance
Using the `icare-extraction-client` branch, running the CSV Client should result in fewer log statements and make the output easier to understand but still give a summary of what is happening.

I also have screenshots of the old logs and the new logs, so if you're interested in a side by side comparison, let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,6 @@ const cancerDiseaseStatusExtractor = new CSVCancerDiseaseStatusExtractor('path-t
 (async () => {
   logger.info('Extracting primary disease status data');
   const primaryDiseaseStatusBundle = await cancerDiseaseStatusExtractor.get({ mrn: 'some-mrn' });
-  logger.info(JSON.stringify(primaryDiseaseStatusBundle));
 })();
 ```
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -39,7 +39,6 @@ const cancerDiseaseStatusExtractor = new CSVCancerDiseaseStatusExtractor('path-t
 (async () => {
   logger.info('Extracting primary disease status data');
   const primaryDiseaseStatusBundle = await cancerDiseaseStatusExtractor.get({ mrn: 'some-mrn' });
-  logger.info(JSON.stringify(primaryDiseaseStatusBundle));
```
jafeltra, line 42: I took this statement out also because our usual procedure is to not log bundles in case there is anything sensitive, so I figured that the documentation should match that.

PR #23, opened 2020-07-31
Update date formats

jafeltra:
# Summary
Updates documentation of date formats. There is still a test that mocks out the MM/DD/YY format and then a bundle that compares to the YYYY-MM-DD format so the actual support for reformatting dates is still tested, and our documentation aligns with the preferred format.


Dtphelan1:
DP1: Would you also mind updating these dates that are out of date (ba dum tiss). While we are here we might as well.
image

Dtphelan1:
This looks good to me! New tests pass, no other instances I could find of the date format.

PR #24, opened 2020-08-03
Expand Patient CSV to handle more values

Dtphelan1:
# Summary
Expand the Patient CSV Extractor to support new fields, including: 
- dateOfBirth,
- race,
- ethnicity,
- language,
- address,
- city,
- state,
- zip

## New behavior
CSV's can have those values and they will be exported into FHIR templates, but templates will still render if some of that data is missing.

## Code changes
- Updated the docs file for patient CSV format
- Added new utils in `patientUtils` for translating ethnicity and race codes
- Updated the `Patient.ejs` template to support these new optional fields
- Updated the CSVPatientExtractor to support these new fields, adding explicit requirement of some fields based on the Spec document provided by Rob in #shr-steam 
- Updated comments to be accurate in `diseaseStatusUtils`
- Created new tests and updated tests where needed

# Testing guidance
- Ensure all tests pass
- Try connecting this with the ICARE-ext-cli and confirm that changing the patient data passed into the csv extractor produces expected output/.

### DIFF
--- a/docs/patient.csv
+++ b/docs/patient.csv
@@ -1,3 +1,3 @@
-mrn,family,given,gender,dateOfBirth,race,ethnicity,language,address,city,state,zip
-pat-mrn-1,last,first,female,1991-06-21,ASKU,2186-5,no,2 West Side Rd,Malden,MA,02186
-pat-mrn-2,final,preceding,male,1986-11-08,2106-3,2186-5,en-US,1 Main street,Brooklyn,NY,11201
+mrn,family,given,gender,birthSex,dateOfBirth,race,ethnicity,language,address,city,state,zip
+pat-mrn-1,last,first,female,female,1991-06-21,ASKU,2186-5,no,2 West Side Rd,Malden,MA,02186
+pat-mrn-2,final,preceding,male,male,1986-11-08,2106-3,2186-5,en-US,1 Main street,Brooklyn,NY,11201
--- a/docs/patient.csv
+++ b/docs/patient.csv
@@ -1,3 +1,3 @@
-mrn,family,given,gender,birthSex,dateOfBirth,race,ethnicity,language,address,city,state,zip
+mrn,family,givenName,genderName,birthSex,dateOfBirth,race,ethnicity,language,addressLine,city,state,zip
 pat-mrn-1,last,first,female,female,1991-06-21,ASKU,2186-5,no,2 West Side Rd,Malden,MA,02186
 pat-mrn-2,final,preceding,male,male,1986-11-08,2106-3,2186-5,en-US,1 Main street,Brooklyn,NY,11201
--- a/docs/patient.csv
+++ b/docs/patient.csv
@@ -1,3 +1,3 @@
-mrn,family,given,gender
-pat-mrn-1,last,first,female
-pat-mrn-2,final,preceding,male
+mrn,family,given,gender,dateOfBirth,race,ethnicity,language,address,city,state,zip
+pat-mrn-1,last,first,female,1991-06-21,ASKU,2186-5,no,2 West Side Rd,Malden,MA,02186
+pat-mrn-2,final,preceding,male,1986-11-08,2106-3,2186-5,en-US,1 Main street,Brooklyn,NY,11201
--- a/docs/patient.csv
+++ b/docs/patient.csv
@@ -1,3 +1,3 @@
-mrn,family,givenName,genderName,birthsex,dateOfBirth,race,ethnicity,language,addressLine,city,state,zip
+mrn,familyName,givenName,gender,birthsex,dateOfBirth,race,ethnicity,language,addressLine,city,state,zip
 pat-mrn-1,last,first,female,female,1991-06-21,ASKU,2186-5,no,2 West Side Rd,Malden,MA,02186
 pat-mrn-2,final,preceding,male,male,1986-11-08,2106-3,2186-5,en-US,1 Main street,Brooklyn,NY,11201
--- a/docs/patient.csv
+++ b/docs/patient.csv
@@ -1,3 +1,3 @@
-mrn,family,givenName,genderName,birthSex,dateOfBirth,race,ethnicity,language,addressLine,city,state,zip
+mrn,family,givenName,genderName,birthsex,dateOfBirth,race,ethnicity,language,addressLine,city,state,zip
 pat-mrn-1,last,first,female,female,1991-06-21,ASKU,2186-5,no,2 West Side Rd,Malden,MA,02186
 pat-mrn-2,final,preceding,male,male,1986-11-08,2106-3,2186-5,en-US,1 Main street,Brooklyn,NY,11201
--- a/src/extractors/CSVPatientExtractor.js
+++ b/src/extractors/CSVPatientExtractor.js
@@ -2,13 +2,16 @@ const path = require('path');
 const { CSVModule } = require('../modules');
 const { generateMcodeResources } = require('../helpers/ejsUtils');
 const { Extractor } = require('./Extractor');
+const { getEthnicityDisplay,
+  getRaceCodesystem,
+  getRaceDisplay } = require('../helpers/patientUtils');
 const logger = require('../helpers/logger');
 
 function joinAndReformatData(patientData) {
   logger.debug('Reformatting patient data from CSV into template format');
   // No join needed, just a reformatting
   const {
-    mrn, family, given, gender, dateOfBirth, race, ethnicity, language, address, city, state, zip,
+    mrn, family, given, gender, birthSex, dateOfBirth, race, ethnicity, language, address, city, state, zip,
   } = patientData;
 
   if (!mrn || !family || !given || !gender) {
@@ -21,14 +24,18 @@ function joinAndReformatData(patientData) {
     family,
     given,
     gender,
+    birthSex,
     dateOfBirth,
-    race,
-    ethnicity,
     language,
     address,
     city,
     state,
     zip,
+    raceCode: race,
+    raceCodesystem: getRaceCodesystem(race),
+    raceText: getRaceDisplay(race),
+    ethnicityCode: ethnicity,
+    ethnicityText: getEthnicityDisplay(ethnicity),
   };
 }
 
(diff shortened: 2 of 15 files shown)


src/templates/Patient.ejs
```diff
@@ -1,5 +1,5 @@
 <%# Patient Template %>
-<%# Based on https://mcodeinitiative.github.io/StructureDefinition-obf-Patient.html %>
+<%# Based on http://hl7.org/fhir/us/mcode/StructureDefinition-mcode-cancer-patient.html %>
 <%# Official url: http://hl7.org/fhir/us/mcode/StructureDefinition/obf-Patient %>
```
jafeltra, line 3: I'm sorry for more annoying comments. I think the official URL is `http://hl7.org/fhir/us/mcode/StructureDefinition/mcode-cancer-patient` now.

src/extractors/CSVPatientExtractor.js
```diff
@@ -7,14 +7,28 @@ const logger = require('../helpers/logger');
...
 
   return {
     id: mrn,
     mrn,
     family,
     given,
     gender,
+    dateOfBirth,
+    race,
+    ethnicity,
```
jafeltra, line 26: Should we be using the `getEthnicityDisplay`, `getRaceCodesystem`, `getRaceDisplay` functions you added to template the additional information here? I think right now, the race and ethnicity provided in the CSV will never get templated since the template checks specifically for the text/code/system values.

src/templates/Patient.ejs
```diff
@@ -8,33 +8,102 @@
...
+  ],
+  "name": [
+      {
+          "text": "<%- `${given} ${family}` %>",
+          "family": "<%- family %>",
+          "given": [ <%- given.split(' ').map(e => `"${e}"`) %> ]
+      }
+  ],
+  "gender": "<%- gender %>"<% if(locals.dateOfBirth && dateOfBirth) { %>,
+  "birthDate": "<%- dateOfBirth%>"<% } %><% if (locals.address && address && locals.city && city && locals.state && state && locals.zip && zip) { %>,
```
jafeltra, line 51: Do you think this condition check could if at least one of `address`, `city`, `stat`, or `zip` is defined so that we can template the pieces that we have? It doesn't seem like all are required if one is present (but obviously I've been fooled by similar things in FHIR before).

src/extractors/CSVPatientExtractor.js
```diff
@@ -7,14 +7,28 @@ const logger = require('../helpers/logger');
 function joinAndReformatData(patientData) {
   logger.info('Reformatting patient data from CSV into template format');
   // No join needed, just a reformatting
-  const { mrn, family, given, gender } = patientData;
+  const {
+    mrn, family, given, gender, dateOfBirth, race, ethnicity, language, address, city, state, zip,
```
jafeltra, line 11: It looks like in Rob's spreadsheet there was a column for `sex`. Is there a reason you didn't include it here?

src/helpers/diseaseStatusUtils.js
```diff
@@ -67,7 +58,7 @@ function getDiseaseStatusEvidenceCode(text) {
 
 /**
  * Converts code in mCODE's CancerDiseaseStatusEvidenceTypeVS to Text Value
- * @param text, limited to No evidence of disease, Responding, Stable, Progressing, or not evaluated
+ * @param text, limited to codes in the evidenceTextToCodeLookup above
  * @return {code} corresponding code from mCODE's CancerDiseaseStatusEvidenceTypeVS
```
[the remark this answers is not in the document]
Dtphelan1: I think actually that this should be {string} actually. I was the [reviewing this page](https://jsdoc.app/tags-param.html) on JS Doc syntax as a part of later comments and based on that I think the comment should actually be: 
- @param {string} text - input text limited to blahblahblah
- @return {string} A corresponding code blahblahblah

Does that look right to you?

src/helpers/diseaseStatusUtils.js
```diff
@@ -49,7 +40,7 @@ function getDiseaseStatusCode(text) {
 
 /**
  * Converts code in mCODE's ConditionStatusTrendVS to Text Value
- * @param text, limited to No evidence of disease, Responding, Stable, Progressing, or not evaluated
+ * @param text, limited to codes in the diseaseStatusTextToCodeLookup above
  * @return {code} corresponding code from mCODE's ConditionStatusTrendVS
```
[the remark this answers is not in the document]
Dtphelan1: See below on JA2.

src/templates/Patient.ejs
Dtphelan1: That's good. I can fix.

src/templates/Patient.ejs
Dtphelan1: I don't know about this. We should chat.

src/extractors/CSVPatientExtractor.js
Dtphelan1: I will add `birthsex`

src/extractors/CSVPatientExtractor.js
Dtphelan1: yep...

src/helpers/diseaseStatusUtils.js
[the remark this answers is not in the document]
Dtphelan1: Sorry for this one it's `code`, not `text`

src/templates/Patient.ejs
Dtphelan1: Should be fixed.

src/extractors/CSVPatientExtractor.js
Dtphelan1: Should be added and fixtures updated. Let me know if I missed anything. @jafeltra

src/extractors/CSVPatientExtractor.js
Dtphelan1: Should be using these now. LEt me know if I missed anything. @jafeltra

src/templates/Patient.ejs
Dtphelan1: We talked. I'll address this shortly.

Dtphelan1: Should be addressed.

src/extractors/CSVPatientExtractor.js
jafeltra: Looks good!

src/templates/Patient.ejs
jafeltra: Looks good to me!

test/extractors/fixtures/csv-patient-module-response.json
```diff
@@ -3,6 +3,18 @@
...
+    "gender": "male",
+    "birthSex": "male",
+    "dateOfBirth":"1994-08-24",
+    "language": "en",
+    "address": "57 Adams St",
+    "city": "New Rochelle",
+    "state": "NY",
+    "zip": "10801",
+    "raceCodesystem": "http://some.codesystem.com",
+    "raceCode": "29178",
```
jafeltra, line 15: I'm not exactly sure this mocked response is working as we'd expect. I think coming back from the CSV module, we'd expect to just have `race` and `ethnicity` and the extractor will then try to get the codesystem and display text. Do you think it makes sense to update these values here?

That also made me realize that if we just use the code values you have here for race and ethnicity, the extensions won't be templated when we use this in the patient extractor test because the util functions won't be able to find corresponding text/codesystems for them. As long as that's the expected behavior, do you think we should use supported codes here so that the extension gets added in the test? Or should we somehow document that this is specifically testing that unsupported codes don't get templated?

Dtphelan1: Good catch. This should be fixed.

Dtphelan1: Longer discussion: Definitely shouldn't have had the text and codesystem values coming back from the CSV service. And as you've pointed out, the code values need to be valid in order have their resultant display and codesystem values be meaningful. I've updated them accordingly to use expected values. I believe the `patientUtils` test file checks how they handle unexpected values. We can talk about how that might live in documentation if you'd like!

Dtphelan1:
I'm good updating our names to align more with Rob's. I have more faith in our ability to fix things to align with the sheet changes than I do have others to update to a new CSV format. No avoiding that with Zip, but we can handle the GivenName change ourselves.

Dtphelan1:
In going through  and updating givenName and familyName, I've also updated addressLine to match the CSV format. I feel strongly that there are important reasons we should align `sex` in our spreadsheet to `birthSex` and not the other way around, so I will leave our use of `birthSex` as is. Happy to discuss my reasoning. Great catches all around Jules. Thanks for the thorough review.

jafeltra:
Cool! Yeah I completely agree on `birthSex` and we need to add `zip` anyway. Good catch on the `addressLine`. And let me know if you wanted to talk at all about the race/ethnicity codes more, but I think those are good.

jafeltra:
This will need a rebase before it gets merged though.

Dtphelan1:
Should be fixed!

docs/patient.csv
```diff
@@ -1,3 +1,3 @@
-mrn,family,given,gender
-pat-mrn-1,last,first,female
-pat-mrn-2,final,preceding,male
+mrn,family,givenName,genderName,birthsex,dateOfBirth,race,ethnicity,language,addressLine,city,state,zip
```
danlee1025, line 1: [DL1] should we update the headers here? `family` -> `familyName`, `genderName` -> `gender`

src/extractors/CSVPatientExtractor.js
```diff
@@ -2,19 +2,40 @@ const path = require('path');
...
 function joinAndReformatData(patientData) {
   logger.debug('Reformatting patient data from CSV into template format');
   // No join needed, just a reformatting
-  const { mrn, family, given, gender } = patientData;
+  const {
+    mrn, familyName, givenName, gender, birthsex, dateOfBirth, race, ethnicity, language, addressLine, city, state, zip,
+  } = patientData;
+
+  if (!mrn || !familyName || !givenName || !gender) {
+    throw Error('Missing required field for Patient CSV Extraction. Required values include: mrn, familyName, given, gender');
```
danlee1025, line 18: [DL2] Update error message: `given` -> `givenName`

docs/patient.csv
Dtphelan1: dang great point.

src/extractors/CSVPatientExtractor.js
Dtphelan1: Dang great point.

PR #27, opened 2020-08-20
Update to return a collection bundle

kjmahalingam:
# Summary

This PR and its accompanying PRs update the BaseFHIRExtractor to return a 'collection' bundle instead of a 'searchset' bundle like before.

## New behavior

Changes as described above.

## Code changes

Small changes to the BaseFHIRExtractor, updating the return value.

# Testing guidance

Run the lints and tests of this and the other PRs, and run this linked with the other repos, and make sure it works! There should be no functional change at all.


kjmahalingam:
One review here is sufficient.

PR #29, opened 2020-08-24
Added BaseClient

danlee1025:
Added `BaseClient` to this repo so that it can be used by extraction clients that import this repo.

Newly created [base-icare-extraction-client](https://github.com/mcode/base-icare-extraction-client) extends this class.


Dtphelan1:
This looks good to me as well. 👍

PR #30, opened 2020-09-02
Conditions new fields

julianxcarter:
# Summary
Added support for several new conditions fields in the CSV extractor.
## New behavior
The condition extractor will now extract attributes for displayName, category, dateOfDiagnosis, clinicalStatus, verificationStatus, bodySite, laterality, and histology when reading a CSV. 
## Code changes

1. Added new example fields in docs/condition.csv 
2. Updated the CSVConditionExtractor to throw an error if mandatory elements are not included and also to read in the new attributes.
3. Updated the condition template to structure the new elements into a valid condition resource.
4. Updated the mock bundle response at test/extractors/fixtures/csv-condition-bundle.json to contain the new elements.
5. Updated the mock module response at test/extractors/fixtures/csv-condition-module-response.json to contain the new elements.
6. Updated the mock condition resource at test/templates/fixtures/condition-resource.json to contain the new elements.
7. Update the CSV condition template test to include the new elements.

# Testing guidance
Should be run with the updated version of the base-icare-extraction-client to ensure that it's getting fed a proper example csv.


src/extractors/CSVConditionExtractor.js
```diff
@@ -3,12 +3,18 @@ const { CSVModule } = require('../modules');
...
 const logger = require('../helpers/logger');
+const { formatDateTime } = require('../helpers/dateUtils');
 
 // Formats data to be passed into template-friendly format
 function formatData(conditionData) {
   logger.debug('Reformatting condition data from CSV into template format');
   return conditionData.map((data) => {
-    const { mrn, conditionId, codeSystem, code } = data;
+    if (!(data.conditionId.trim() && data.mrn.trim() && data.codeSystem.trim() && data.code.trim() && data.category.trim())) {
+      throw new Error('The condition is missing an expected attribute. Condition id, subject, code system, code, and category are all required.');
```
kjmahalingam, line 13: A few things about this line.

DM1: Seems that this warning is missing a `mrn` as an expected attribute despite it being tested. Could you add that to the warning?
DM2: Is there a need for the `.trim()` here (as well as other locations throughout this file)? If you're just including it to be safe, I think it's not necessary; in a CSV file we can expect that there won't be leading or trailing spaces between the commas. If there's a specific reason, happy to hear it! If we can remove them here, could we also remove them in the other locations you have them through this file?
DM3: If we can remove the trims as per DM2, you can move this below the next set of lines where you destructure the data object. That way you don't have to say `data.` in front of every field.

src/extractors/CSVConditionExtractor.js
```diff
@@ -18,6 +24,37 @@ function formatData(conditionData) {
       code: {
         code,
         system: codeSystem,
+        display: displayName,
+      },
+      category: category.split('|').map((categoryCode) => ({
+        system: 'http://terminology.hl7.org/CodeSystem/condition-category',
+        code: categoryCode,
+      })),
+      dateOfDiagnosis: !dateOfDiagnosis.trim() ? null : {
```
kjmahalingam, line 33: As mentioned in DM2: I think that we can probably get read of all of these `.trim()`s.

src/extractors/CSVConditionExtractor.js
julianxcarter: DM1: `mrn` has been added to the warning in the last commit. Likewise, `subject` has been removed.
DM2: I was totally being overcautious! I wasn't sure if a string containing just a space like `' '` would throw off the if-statement. All of the `.trim()`'s have been removed in the most recent commit.
DM3: This has been applied in the most recent commit. The error message is now placed below the `const` statement, and the if-statement now checks the newly created variables instead of using the `data.` prefix.

src/extractors/CSVConditionExtractor.js
julianxcarter: They've all been removed in the most recent commit! Thank you for the feedback :).

Dtphelan1:
DP1 - which is definitely not on you but is something I've only now noticed: We really should have three different tests for the templates. One with a maximal set of data; one with the minimal set of data; and one that is missing required data. We don't do this today, but we should because when I added such a test it showed some issues with the current template 😢  

Action Item: Could you add a test that renders the Condition template with the following set of required data elements, and fix the templates to handle any errors that get thrown. Happy to talk more about errors. 
```
const CONDITION_MINIMAL_DATA = {
  // Only include 'subject', 'conditionId', 'code', 'codesystem', and 'category' fields which are required
  id: 'example-id',
  subject: {
    id: 'example-subject-id',
  },
  code: {
    system: 'example-system',
    code: 'example-code',
    display: 'exampleDisplayName',
  },
  category: [
    {
      system: 'example-system',
      code: 'example-code',
    },
  ],
};
```

Dtphelan1:
This looks good to me! Thanks for addressing all my comments so swiftly!

Dtphelan1 (approved):

src/templates/Condition.ejs
```diff
@@ -4,59 +4,132 @@
...
+          }
         }
-      ]
-    },
-    {
+      ], <% } %>
       "coding": [
         {
-          "system": "http://snomed.info/sct",
-          "code": "64572001"
```
jafeltra, line 49: I think we want to keep this hardcoded coding in the array in addition to the one that gets templated. This coding is a little weird, but for now, we want it to be on all conditions along with any other category coding that is provided. It technically should only be on cancer conditions (you can see it on the Cancer Condition Parent profile [here](http://hl7.org/fhir/us/mcode/StructureDefinition-mcode-cancer-condition-parent.html)) and we have a [Jira task](https://jira.mitre.org/browse/STEAM-362) to update to only include it on those cancer conditions, but for now, I think we want it to always be present. This shouldn't be too difficult of a change!

jafeltra (commented):
This looks really awesome! I'm still having issues trying to actually validate any FHIR resource (it's a problem with my tools, not your code), so I haven't been able to do that, but looking at the resource, it seems correct. I'm going to try one last thing to try validating, so I'll let you know how that goes. I only had one minor comment about the code, which is below.

src/extractors/CSVConditionExtractor.js
```diff
@@ -18,6 +24,37 @@ function formatData(conditionData) {
...
+      clinicalStatus: !clinicalStatus ? null : {
+        system: 'http://terminology.hl7.org/CodeSystem/condition-clinical',
+        code: clinicalStatus,
+      },
+      verificationStatus: !verificationStatus ? null : {
+        system: 'http://terminology.hl7.org/CodeSystem/condition-ver-status',
+        code: verificationStatus,
+      },
+      bodySite: !bodySite ? null : bodySite.split('|').map((site) => ({
+        system: 'http://snomed.info/sct/',
```
jafeltra, line 46: JA2: Super minor, but the tool I was trying to use to validate was complaining about the trailing `/` here. I think if you remove it, it stops compaining.

jafeltra (commented):
I tried my best to validate the resulting bundle and only noticed one, incredibly minor thing it complained about with these new changes, which is awesome!

src/templates/Condition.ejs
julianxcarter: JA1: Gotcha! I see in the profile that you linked where the mandatory coding is specified. I just added that back with this most recent commit.

src/extractors/CSVConditionExtractor.js
julianxcarter: JA2: I never would've caught this! I removed the trailing backslash in the most recent commit. Nice catch.

jafeltra: I never would have caught it if the validator hadn't caught if for me!

jafeltra (approved):
This looks great! I didn't notice anything else besides those two things you've fixed. Awesome work!

PR #32, opened 2020-09-08
New fields added to cancer disease status

julianxcarter:
# Summary
Added support for an additional `observationStatus` field in the CancerDiseaseStatus extractor.
## New behavior
The extractor can now extract attributes for `observationStatus` and `diseaseStatusText`. It will no longer be able to extract a value for `dateRecorded`, per the example csv file. 
## Code changes
1. Added new example fields in `docs/cancer-disease-status.csv`.
2. Updated the Cancer Disease Status extractor to throw an error if `observationStatus` is omitted given that it is a mandatory attribute. Also updated the extractor to read in a display value for `diseaseStatus`, falling back to a lookup with the `diseaseStatus` code by way of the helper functions in `src/helpers/diseaseStatusUtils.js` if a display value is not included.
3. Updated the diseaseStatusText values in `src/helpers/diseaseStatusUtils.js`. They may have been updated since the extractor was first built, but the ones included in the file no longer matched up with the values on the [reference site](http://standardhealthrecord.org/guides/icare/mapping_guidance.html).
4. Updated all test cases to correlate with additions made to the extractor. Also, added a minimal viable test case that only passes in mandatory values to the template.
# Testing guidance
Should be tested with the updated base-icare-extraction-client.

### DIFF
--- a/src/helpers/diseaseStatusUtils.js
+++ b/src/helpers/diseaseStatusUtils.js
@@ -10,11 +10,11 @@ function mEpochToDate(date) {
 // Code mapping is based on http://standardhealthrecord.org/guides/icare/mapping_guidance.html
 // specifically using lowercase versions of the text specified by ICARE for status
 const diseaseStatusTextToCodeLookup = {
-  'no evidence of disease': 260415000,
-  responding: 268910001,
-  stable: 359746009,
-  progressing: 271299001,
-  'not evaluated': 709137006,
+  'Not detected (qualifier)': 260415000,
+  'Patient condition improved (finding)': 268910001,
+  'Patient\'s condition stable (finding)': 359746009,
+  'Patient\'s condition worsened (finding)': 271299001,
+  'Patient condition undetermined (finding)': 709137006,
 };
 const diseaseStatusCodeToTextLookup = invertObject(diseaseStatusTextToCodeLookup);
 
--- a/src/helpers/diseaseStatusUtils.js
+++ b/src/helpers/diseaseStatusUtils.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-else-return */
 const moment = require('moment');
 const { invertObject } = require('./helperUtils');
 
@@ -45,11 +44,12 @@ const evidenceCodeToTextLookup = invertObject(evidenceTextToCodeLookup);
  * @param {string} text, limited to 'no evidence of disease', Responding, Stable, Progressing, or 'not evaluated'
  * @return {code} corresponding DiseaseStatus code
  */
-function getDiseaseStatusCode(text, implementation = 'icare') {
-  if (implementation === 'icare') {
-    return icareDiseaseStatusTextToCodeLookup[text];
-  } else {
-    return currentDiseaseStatusTextToCodeLookup[text];
+function getDiseaseStatusCode(text, implementation) {
+  switch (implementation) {
+    case 'icare':
+      return icareDiseaseStatusTextToCodeLookup[text];
+    default:
+      return currentDiseaseStatusTextToCodeLookup[text];
   }
 }
 
@@ -58,11 +58,12 @@ function getDiseaseStatusCode(text, implementation = 'icare') {
  * @param {string} code - limited to codes in the diseaseStatusTextToCodeLookup above
  * @return {string} corresponding DiseaseStatus display text
  */
-function getDiseaseStatusDisplay(code, implementation = 'icare') {
-  if (implementation === 'icare') {
-    return icareDiseaseStatusCodeToTextLookup[code];
-  } else {
-    return currentDiseaseStatusCodeToTextLookup[code];
+function getDiseaseStatusDisplay(code, implementation) {
+  switch (implementation) {
+    case 'icare':
+      return icareDiseaseStatusCodeToTextLookup[code];
+    default:
+      return currentDiseaseStatusCodeToTextLookup[code];
   }
 }
 
--- a/src/helpers/diseaseStatusUtils.js
+++ b/src/helpers/diseaseStatusUtils.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-else-return */
 const moment = require('moment');
 const { invertObject } = require('./helperUtils');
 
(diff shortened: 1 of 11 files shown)


src/templates/CancerDiseaseStatus.ejs
```diff
@@ -30,7 +30,22 @@
     "profile": [
       "http://hl7.org/fhir/us/mcode/StructureDefinition/mcode-cancer-disease-status"
     ]
-  },
+  }, <% if (typeof evidence !== 'undefined' && evidence && evidence.length > 0) { %>
```
kjmahalingam, line 33: DM1: `undefined` in JavaScript evaluates to `false` in checks like this, so you don't need to actually add this `typeof evidence !== 'undefined'` here by my understanding. 

As a separate question, is there any change you made to this file (aside from DM1 above) that wasn't just reordering things in here? I can't see one so just want to confirm!

kjmahalingam (changes requested):
This is awesome! I have one quick comment and question inline, but otherwise this looks like awesome work that is in line with the required documentation.

src/templates/CancerDiseaseStatus.ejs
julianxcarter: DM1: You're absolutely right! I began adding in `typeof <variable> !== 'undefined'` as a habit last week to insure that a reference error isn't thrown if any of the variables haven't been declared. All of the extractors do declare variables for each of the possible attributes (and just make them `null` if the attribute wasn't included on the CSV), but Dylan P and I thought it would be good to have that extra level of protection. It can be dropped though since we wouldn't ever run into that problem with the extractors as they're currently designed.

Also, there were no other changes made to the template, I just reordered some things. Everything was already included when I began :).

kjmahalingam: DM1: Thanks for the clarification! If @Dtphelan1 is also fine with this change, that's okay with me; it's ultimately harmless. :) I do, however, think that you can just check for `evidence !== undefined` instead of doing the `typeof` check though.

julianxcarter: DM1: Gotcha! I think I'll go ahead and remove it since it's not likely to be an error that we encounter.

kjmahalingam (approved):
Nice work!

kjmahalingam (changes requested):
Err...sorry if I mislead you, but I guess some tests are actually now failing due to ReferenceErrors. :(

I think the check mentioned in DM1 can probably be re-implemented, since that resolves the issue. Sorry again!

kjmahalingam (approved):
Alright, just ran and tested it again, and I *actually definitively approve* now. Good work!

Dtphelan1:
Just for record keeping: Julian and team talked today about how to ensure that the CDS extractors can support both the ICARE specific valueSystem codes as well as the generic mCODE codes. We've come up with a reasonable approach that is low lift, and will help realize the overall task. Looking forward to reviewing those addl. changes come next week! 👍

kjmahalingam:
After the changes, still approved here! A couple small notes on the other PR, but great work. :)

src/helpers/diseaseStatusUtils.js
```diff
@@ -34,17 +45,25 @@ const evidenceCodeToTextLookup = invertObject(evidenceTextToCodeLookup);
...
 }
 
 /**
  * Converts code in mCODE's ConditionStatusTrendVS to Text Value
  * @param {string} code - limited to codes in the diseaseStatusTextToCodeLookup above
  * @return {string} corresponding DiseaseStatus display text
  */
-function getDiseaseStatusDisplay(code) {
-  return diseaseStatusCodeToTextLookup[code];
+function getDiseaseStatusDisplay(code, implementation = 'icare') {
```
Dtphelan1, line 61: DP - If we are going to use a default arg value here, we probably want it to be "mcode", but given how the if-else statement works I don't think we need a default value at all. This also applies to the block of code above in the `getDiseaseStatusCode` function. 😄

src/helpers/diseaseStatusUtils.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-else-return */
```
Dtphelan1, line 1: DP - If we want to avoid making es-lint unhappy, we might want to use a switch statement instead of an if-else block in the disease status lookup functions.

test/helpers/diseaseStatusUtils.test.js
```diff
@@ -8,12 +8,12 @@ const {
 
 // Code mapping is based on http://standardhealthrecord.org/guides/icare/mapping_guidance.html
 // specifically using lowercase versions of the text specified by ICARE for status
-const diseaseStatusTextToCodeLookup = {
-  'no evidence of disease': 260415000,
-  responding: 268910001,
-  stable: 359746009,
-  progressing: 271299001,
-  'not evaluated': 709137006,
+const currentDiseaseStatusTextToCodeLookup = {
```
Dtphelan1, line 11: DP - probably want two local lookup tables here now, right? And we want to make sure that changing the implementation flag will change the output? We should add a function for that too.

src/helpers/diseaseStatusUtils.js
julianxcarter: DP - Switched to a switch statement in the most recent commit and made the default case reference the current lookup table, which addressed the preview comment as well. Es-lint is now pleased :).

test/helpers/diseaseStatusUtils.test.js
julianxcarter: DP - Added a new lookup table and new functional test in the most recent commit!

test/helpers/diseaseStatusUtils.test.js
```diff
@@ -26,18 +36,30 @@ const evidenceTextToCodeLookup = {
 };
 
 describe('diseaseStatusUtils', () => {
-  test('getDiseaseStatusCode,', () => {
+  test('getMcodeDiseaseStatusCode,', () => {
     Object.keys(currentDiseaseStatusTextToCodeLookup).forEach((dsText) => {
       const dsCode = currentDiseaseStatusTextToCodeLookup[dsText];
       expect(getDiseaseStatusCode(dsText, 'mcode')).toEqual(dsCode);
     });
```
Dtphelan1: DP - Super small, but in the body of this loop could we add a second expect statement - `expect(getDiseaseStatusCode(dsText)).toEqual(dsCode);`  - that ensures our default case works when there is no argument supplied?

Dtphelan1: Ditto the the `getMcodeDiseaseStatusDisplay` test below

Dtphelan1:
After those changes, I believe we should be all set to approve! 🎉

kjmahalingam (approved):
I’m still good to approve, nice work!

test/helpers/diseaseStatusUtils.test.js
Dtphelan1: @julianxcarter in case this notification slipped through your inbox last night 👍

julianxcarter: DP - Just pushed up a commit with these changes! 👍🏽

Dtphelan1 (approved):
Looks good - set to merge after you rebase! 😄

PR #34, opened 2020-09-16
Add dateRecorded column to Cancer Disease Status

jafeltra:
I realized that the `dateRecorded` column was removed from the CancerDiseaseStatus CSV sheet. We had added this column to reflect the date the entry was added to the CSV sheet and use it to filter out data during extraction. However, since this doesn't map onto an element in the FHIR resource, it isn't reflected in Rob's spreadsheet. I can update the spreadsheet as well, as long as everyone agrees we should still have this column.


kjmahalingam (approved):
💯 🥇

Dtphelan1:
Looks good to me - confirmed that we still use this field appropriately in the CSVModule as well. WooHoo!!!

Dtphelan1 (approved):

PR #36, opened 2020-09-23
Template revamp

Dtphelan1:
# Summary
Begins the revamp of our template approach, backing away from EJS and moving towards home-rolled JavaScript template functions that receive template data and produce JSONified FHIR resources. This task starts with the Patient resource, adding fundamental snippets and decorators for the rapid and reusable creation of future templates. Additionally, the approach taken here is one that aims to be progressive; the previous approach for templating with EJS is still used for non-updated templates. As we update additional templates, we need to make minor changes to our `EJSUtils` file until we are finished with the EJS approach, at which point we'll need to remove that file and update the `templateUtils` accordingly. 

# Detailed Notes
Below is a list of detailed notes on the 
- `EJSUtils` has a list of updated templates, as well as an update to the FHIRTemplateLookup table to point at the newly created Patient TemplateFunction. 
- `templateUtils` (where some of the old `EJSUtils` functionality will eventually live) contains several decorators for templateFunctions. These decorators enable the separation of requirements for which arguments need to be present from the nature of the template itself. 
- Various places use the below motif for adding a property and value only-if the provided value is non-falsy. This is helpful in avoiding empty properties on the resultant JSON object, though there are other ways we could trim these properties if we wanted to explore them. Note also that `...null` evaluates without throwing an error, and produces no new property.
```javascript
{
    address: [
      {
        ...(addressLine && {
          line: [
            addressLine,
          ],
        }),
        ...(city && { city }),
        ...(state && { state }),
        ...(zip && { postalCode: zip }),
        country: 'US',
      },
    ],
  };
```

# Making New Templates
Making a new template involves the following steps: 
- Create a new `.js` file responsible for taking incoming data as described by the current `.ejs` template and producing the JSONified FHIR Resources.
- Uses the decorators, `...(property && {property})` pattern and any number of smaller functions you create to render the various subcomponents of the Patient resource, 
- Add tests for any new sub-functions, and update previous template tests to ensure that it tests the new templateFunction
- Update the `EJSUtils` file to ensure that the `mcodeProfileID` is added to the new templates list, ensuring that the render function uses the appropriate render logic for that template. 

# Testing guidance
- Ensure all tests are passing; 
- Ensure that there aren't any other tests that need updating;
- Review the current snippets and comment on any ones we need to add;

Happy Reviewing!


src/helpers/templateUtils.js
```diff
@@ -0,0 +1,56 @@
...
+  return (...args) => {
+    if (_.some(args, (arg) => _.isUndefined(arg) || _.isNull(arg))) {
+      return null;
+    }
+    return templateFn(...args);
+  };
+}
+
+// A decorator that modifies a template to only render
+// when all of the values on its argumentsObject are null/empty, returning null otherwise
```
kjmahalingam, line 16: Should this comment say when it's *not* null/empty? That's what I'm gathering from the functionality.

src/helpers/templateUtils.js
```diff
@@ -0,0 +1,56 @@
...
+  return (...args) => {
+    if (_.every(args, (arg) => _.isUndefined(arg) || _.isNull(arg))) {
+      return null;
+    }
+    return templateFn(...args);
+  };
+}
+
+// A decorator that modifies a template to only render
+// when some of the values on its argumentsObject are null/empty, returning null otherwise
```
kjmahalingam, line 40: Same comment as above, should this say *not* null/empty?

kjmahalingam (commented):
This is awesome, really good work. I like this methodology a lot, and I like the way you've organized it. I particularly appreciate your efforts to make this easy to modify as we all work on individual templates, as well.

I didn't do a full review, just a pass over the code, so don't count me as one of the approvals. Just wanted to get my thoughts across before I left for my day off, so as not to hold this off at all.

I had a couple very minor questions I made in-line, based on what I think might be a typo in some comments, but other than that this passing review looks great to me!

src/helpers/templateUtils.js
Dtphelan1: uhhh uhhh uhhh.

src/helpers/templateUtils.js
Dtphelan1: uhhh. uhhh uhhh.

Dtphelan1: Addressed. 😄

src/helpers/templateUtils.js
Dtphelan1: Addressed. 😄

test/helpers/templateUtils.test.js
```diff
@@ -0,0 +1,183 @@
+const { ifAllArgs, ifAllArgsObj, ifSomeArgs, ifSomeArgsObj } = require('../../src/helpers/templateUtils.js');
+
+describe('TemplateUtils', () => {
+  describe('ifAllArgs', () => {
+    const fn = (a1, a2) => ({ arg1: a1, arg2: a2 });
+    const ifAllArgsFn = ifAllArgs(fn);
+
+    test('Returns nothing when all args are empty', () => {
+      expect(ifAllArgsFn(undefined, undefined)).toBeNull();
+      expect(ifAllArgsFn(null, null)).toBeNull();
```
danlee1025, line 10: [DL1] Maybe add a test case where the args are a mix of `undefined` and `null` like `expect(ifAllArgsFn(null, undefined)).toBeNull();`

[DL2] Do you think we should add a test case where no args are passed?

test/helpers/templateUtils.test.js
```diff
@@ -0,0 +1,183 @@
...
+      expect(ifAllArgsObjFn({ a1: vNum, a2: undefined })).toBeNull();
+      expect(ifAllArgsObjFn({ a1: undefined, a2: vNum })).toBeNull();
+      expect(ifAllArgsObjFn({ a1: vNum, a2: null })).toBeNull();
+      expect(ifAllArgsObjFn({ a1: null, a2: vNum })).toBeNull();
+      // Permutations with Str
+      expect(ifAllArgsObjFn({ a1: vStr, a2: undefined })).toBeNull();
+      expect(ifAllArgsObjFn({ a1: undefined, a2: vStr })).toBeNull();
+      expect(ifAllArgsObjFn({ a1: vStr, a2: null })).toBeNull();
+      // Permutations with Bool
+      expect(ifAllArgsObjFn({ a1: null, a2: vStr })).toBeNull();
```
danlee1025, line 74: [DL2] I think this line belongs on the line above 😂

test/helpers/templateUtils.test.js
```diff
@@ -0,0 +1,183 @@
...
+      expect(ifSomeArgsFn(null, vStr)).toEqual(fn(null, vStr));
+      // Permutations with Bool
+      const vBool = true;
+      expect(ifSomeArgsFn(vBool, undefined)).toEqual(fn(vBool, undefined));
+      expect(ifSomeArgsFn(undefined, vBool)).toEqual(fn(undefined, vBool));
+      expect(ifSomeArgsFn(vBool, null)).toEqual(fn(vBool, null));
+      expect(ifSomeArgsFn(null, vBool)).toEqual(fn(null, vBool));
+    });
+
+    test('Returns the function output when all args are empty', () => {
```
danlee1025, line 128: [DL3] Did you mean when all args are present here?

test/helpers/templateUtils.test.js
```diff
@@ -0,0 +1,183 @@
...
+      expect(ifSomeArgsObjFn({ a1: null, a2: vStr })).toEqual(objFn({ a1: null, a2: vStr }));
+      // Permutations with Bool
+      const vBool = true;
+      expect(ifSomeArgsObjFn({ a1: vBool, a2: undefined })).toEqual(objFn({ a1: vBool, a2: undefined }));
+      expect(ifSomeArgsObjFn({ a1: undefined, a2: vBool })).toEqual(objFn({ a1: undefined, a2: vBool }));
+      expect(ifSomeArgsObjFn({ a1: vBool, a2: null })).toEqual(objFn({ a1: vBool, a2: null }));
+      expect(ifSomeArgsObjFn({ a1: null, a2: vBool })).toEqual(objFn({ a1: null, a2: vBool }));
+    });
+
+    test('Returns the function output when all args are empty', () => {
```
danlee1025, line 171: [DL4] Same as DL3

src/templates/PatientTemplate.js
```diff
@@ -0,0 +1,150 @@
...
+      {
+        text: `${givenName} ${familyName}`,
+        family: familyName,
+        given: givenName.split(' '),
+      },
+    ],
+  };
+}
+
+function addressTemplate({ addressLine, city, state, zip }) {
```
danlee1025, line 88: [DL5] Should probably add some tests for `addressTemplate` and `nameTemplate` since they seem a little more complicated than the other functions in this file.

test/helpers/templateUtils.test.js
Dtphelan1: Addressed

test/helpers/templateUtils.test.js
Dtphelan1: Addressed

test/helpers/templateUtils.test.js
Dtphelan1: Addressed

test/helpers/templateUtils.test.js
Dtphelan1: Addressed

src/templates/PatientTemplate.js
```diff
@@ -0,0 +1,150 @@
+const { extension, coding } = require('./snippets');
+const { ifAllArgsObj, ifSomeArgsObj } = require('../helpers/templateUtils');
+
+function mrnIdentifierTemplate({ mrn }) {
+  return {
+    identifier: [
+      {
+        type: {
+          coding: [
```
julianxcarter, line 9: Just a question, is there a reason that you use the `coding.js` snippet on the `ValueCoding` elements but not within this array? It seems like that functionality could apply here as well.

Dtphelan1: There is no good reason AFAIK, I just must have missed it. 😆  Thanks for catching this Julian!

src/templates/PatientTemplate.js
Dtphelan1: That would be a good thing to do! 👍

src/templates/PatientTemplate.js
Dtphelan1: Should be resolved, and updated a few other places to also be in alignment with this suggestion. LMK what you think!

test/helpers/templateUtils.test.js
```diff
@@ -0,0 +1,191 @@
...
+      const outputStr = objFn({ a1: vStr, a2: vStr });
+      expect(ifAllArgsObjFn({ a1: vStr, a2: vStr })).toEqual(outputStr);
+      const vBool = true;
+      const outputBool = objFn({ a1: vBool, a2: vBool });
+      expect(ifAllArgsObjFn({ a1: vBool, a2: vBool })).toEqual(outputBool);
+    });
+  });
+
+  describe('ifSomeArgs', () => {
+    const fn = (a1, a2) => `Things to care about include ${a1 && a1}, ${a2 && a2} and of course your own happiness.`;
```
jafeltra, line 103: :)

test/templates/patient.test.js
```diff
@@ -1,11 +1,6 @@
...
-const path = require('path');
 const basicPatient = require('./fixtures/patient-resource.json');
 const maximalPatient = require('./fixtures/maximal-patient-resource.json');
-const { renderTemplate } = require('../../src/helpers/ejsUtils');
-
-const PATIENT_TEMPLATE = fs.readFileSync(path.join(__dirname, '../../src/templates/Patient.ejs'), 'utf8');
-
+const { patientTemplate } = require('../../src/templates/PatientTemplate');
 
 describe('EJS Render Patient', () => {
```
jafeltra, line 5: Super small, but should we have this say `JavaScript Render Patient` or something similar?

jafeltra (commented):
This looks really great! I really like this approach and I think it'll make things a lot easier to manage going forward. I just had a few, pretty minor comments. If you think any are out of scope of this PR or you want to talk about any, let me know!

src/templates/snippets/valueX.js
```diff
@@ -0,0 +1,89 @@
...
+
+// Regex Patterns
+const quantRegex = /^-?(0|[1-9][0-9]*)(\.[0-9]+)?([eE][+-]?[0-9]+)?((\s){1,3}(\S{1,15}))?$/;
+const dateRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|1[0-2])(-(0[1-9]|[1-2][0-9]|3[0-1]))?)?$/;
+const dateTimeRegex = /^\d\d\d\d-\d\d-\d\dT\d\d(:\d\d(:\d\d(-\d\d:\d\d)?)?)?$/;
+const valueTimeRegex = /^([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\.[0-9]+)?$/;
+
+function valueX(value) {
+  switch (typeof value) {
+    case 'number':
```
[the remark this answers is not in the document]
Dtphelan1: That's an excellent point. I'll look at the ValueTypes possible and add at least this one, maybe leave TODO's for the others. Good catch.

test/templates/patient.test.js
Dtphelan1: Yes it should....

src/templates/snippets/valueX.js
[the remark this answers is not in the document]
jafeltra: I just looked in SUSHI and remembered there is also the possibility for unsignedInt and positiveInt as well. So I'm totally okay with just leaving TODOs for all these and keep the case as is for now. In case it is helpful, SUSHI checks the following:

```
type === 'decimal' ||
(type === 'integer' && Number.isInteger(value)) ||
(type === 'unsignedInt' && Number.isInteger(value) && value >= 0) ||
(type === 'positiveInt' && Number.isInteger(value) && value > 0)
```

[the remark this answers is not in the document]
jafeltra: The more I think about this, the more I think we should just wait since this covers all the cases we currently need. We can always add as we need them or have a separate task to build out full valueX functionality.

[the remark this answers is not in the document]
Dtphelan1: Will wait and make a new task

src/templates/PatientTemplate.js
```diff
@@ -0,0 +1,150 @@
...
+      },
+    ],
+    url: 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-ethnicity',
+  };
+}
+
+function usCoreBirthSexExtensionTemplate({ birthsex }) {
+  return {
+    url: 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-birthsex',
+    valueCode: birthsex,
```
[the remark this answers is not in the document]
Dtphelan1: If you're asking why I didn't implement these cases, it's because I just adapted Rob's code and I suppose he didn't. 😆 . We should implement this before this PR is done. Will work on figuring out how to separate one case from the other. This is good feedback.

src/templates/PatientTemplate.js
```diff
@@ -0,0 +1,150 @@
...
+      {
+        url: 'ombCategory',
+        ...valueX({
+          system: 'urn:oid:2.16.840.1.113883.6.238',
+          code: ethnicityCode,
+        }),
+      },
+      {
+        url: 'text',
+        valueString: ethnicityText,
```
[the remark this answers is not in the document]
Dtphelan1: Yes. Doing that now.

src/templates/PatientTemplate.js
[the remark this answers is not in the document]
Dtphelan1: This is a harder question than anticipated. Not sure how we want to proceed here. Will talk tomorrow at standup.

kjmahalingam:
I have done a re-review, and I think this still looks awesome. I agree with the notes made in other comments. Only one other thing I have to add.

DM1: What is the plan if we have a field that is possible to use across templates? For example, I know that the name template and the language template are written in the patient template file right now because those are realistically the only resources where we'd need those, given what we care about right now. But what if we had a field that we wanted to template that could be used across various templates? Would that then go in a snippet? It seems like a little bit of a different case than the way we're using snippets for codings or valueXs right now, so just want to make sure. Additionally, I can see a reason for wanting the smaller templates in the patient template file to *already* be in a separate utils/snippets file, so I wouldn't be opposed to that. Don't necessarily want a code change associated with this comment, just an answer.

Dtphelan1:
DM1 - My thinking was that it would live in the snippets folder, a kind of grabbag of subcomponents to be used all over. If that folder gets out of hand, we can organize it later. But I don't want to force any organization now since I have little idea how I would structure it. My mental workflow has been
1. Move full template into a .js file;
2. Decompose into subcomponents that "make sense" (vagueness here may be a blessing or a curse. Time will tell);
3. For subcomponents that would reasonably be reused, or when in need of a subcomponent known to be used by another template, move them out into a snippet file;

As for wanting smaller templates in the patient file to be in a separate utils/snippet file, I think I want to push back there slightly. I agree with Dan that I might want to make the tests a little more robust to ensure that errors aren't thrown when more minimal sets of data are supplied. But I don't know if we need to get into the business of testing functions that aren't exposed to anyone else. There's a lot here, so happy to talk more at or after standup.

kjmahalingam:
On @Dtphelan1 's response to DM1: I'm perfectly happy with this answer, and therefore once all comments other than this are addressed, I'm good to approve.

src/templates/PatientTemplate.js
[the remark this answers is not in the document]
Dtphelan1: STEAM-376 has been added to support a new approach needed in order to handle this case. Not worrying about it now.

src/templates/PatientTemplate.js
Dtphelan1: I think I cover a more robust series of tests against `addressTemplate` by using a new helper that tests all permutations of optional arguments for not failing. As for `nameTemplate`, I don't think we need to worry about testing all the permutations of nameTemplate since the familyName and givenName are always provided. 

Happy to talk more about this, but I feel like this is in a good state given our current requirements. We confirm that data is rendered as expected in both minimal and maximal cases, and we confirm that errors are avoided in all combinations of possible arguments; if we notice weird output in the future given some odd corner case of arguments, we can create appropriate fixtures and regression tests for that oddity. But I don't think we want to try and create fixtures for what all valid combinations of template arguments will look like. 

</Rant>

Let me know if that makes sense.

jafeltra (approved):
All the JA's look good to me now! From what I have read/understood, all the DMs and DLs look good to me as well, so I feel comfortable approving this PR. But we should wait to merge for the other reviewers to also approve when they feel ready.

kjmahalingam (approved):
All clear to me!

src/templates/PatientTemplate.js
danlee1025: sounds good to me!

danlee1025 (approved):
LGTM, great work!

julianxcarter (approved):
This looks great to me! And I'll also add that in my experience, it was not particularly tedious to replicate in other templates so I'd say it's serving its purpose. 👍🏽

Dtphelan1:
And with that, merging!

PR #38, opened 2020-09-30
Minor updates for ObservationStatus and Study Reference

jafeltra:
# Summary
While I was reviewing the ICARE documentation, I noticed a couple things that the extraction framework needed to update.

## New behavior
- We added in an `observationStatus` field in the cancer disease status CSV, but this should be an optional field in the CSV file. Even though it is a required `1..1` property in FHIR, we can reasonably assume a default value of `final` since this is describing the status of the FHIR resource and not anything to do with an observed piece of patient data (also confirmed this with Rob and others at stand up last week). This PR makes the `observationStatus` field no longer a required field and defaults it to 'final' if it is not provided.

- The way we were referencing the ResearchStudy on the ResearchSubject was incorrect. Most of our references to other resources within the bundle use the `urn:uuid` format, but this was still using an old `ResearchStudy/id` format. The old implementation was not correct because we were using an identifier value rather than the resource's id. In order to use the id of the ResearchStudy in the `urn:uuid` format, we'd need to generate the full ResearchStudy resource in order to generate the id, and then use that id when templating ResearchSubject. Rather than make those bigger changes to the extractor, I updated the reference to properly refer to the identifier of the study. I believe this idea is a "[business identifier](https://www.hl7.org/fhir/resource.html#identifiers)" since it is referencing an underlying, unchanging concept - the particular study. This format of a reference is also shown in the [bundle references example](http://www.hl7.org/fhir/bundle-references.json.html).

## Code changes
Observation status is not required and defaults to final. Support for the new reference format for `study` on the ResearchSubject.

# Testing guidance
Ensure you can run the extraction client without an `observationStatus` provided and ensure the identifier is templated correctly. Ensure tests still pass.

### DIFF
--- a/test/extractors/CSVCancerDiseaseStatusExtractor.test.js
+++ b/test/extractors/CSVCancerDiseaseStatusExtractor.test.js
@@ -25,13 +25,14 @@ const csvModuleSpy = jest.spyOn(csvModule, 'get');
 describe('CSVCancerDiseaseStatusExtractor', () => {
   describe('joinAndReformatData', () => {
     test('should join data appropriately and throw errors when missing required properties', () => {
-      const expectedErrorString = 'DiseaseStatusData missing an expected property: mrn, conditionId, diseaseStatusCode, observationStatus, and dateOfObservation are required.';
+      const expectedErrorString = 'DiseaseStatusData missing an expected property: mrn, conditionId, diseaseStatusCode, and dateOfObservation are required.';
       const localData = _.cloneDeep(exampleCSVDiseaseStatusModuleResponse);
       // Test that valid data works fine
       expect(csvCancerDiseaseStatusExtractor.joinAndReformatData(exampleCSVDiseaseStatusModuleResponse)).toEqual(expect.anything());
 
       // Test all required properties are
       delete localData[0].evidence; // Evidence is not required and will not throw an error
+      delete localData[0].observationStatus; // Observation Status is not required and will not throw an error
       Object.keys(localData[0]).forEach((key) => {
         const clonedData = _.cloneDeep(localData);
         delete clonedData[0][key];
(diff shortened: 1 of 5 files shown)


test/extractors/CSVCancerDiseaseStatusExtractor.test.js
```diff
@@ -25,13 +25,14 @@ const csvModuleSpy = jest.spyOn(csvModule, 'get');
...
     test('should join data appropriately and throw errors when missing required properties', () => {
-      const expectedErrorString = 'DiseaseStatusData missing an expected property: mrn, conditionId, diseaseStatusCode, observationStatus, and dateOfObservation are required.';
+      const expectedErrorString = 'DiseaseStatusData missing an expected property: mrn, conditionId, diseaseStatusCode, and dateOfObservation are required.';
       const localData = _.cloneDeep(exampleCSVDiseaseStatusModuleResponse);
       // Test that valid data works fine
       expect(csvCancerDiseaseStatusExtractor.joinAndReformatData(exampleCSVDiseaseStatusModuleResponse)).toEqual(expect.anything());
 
       // Test all required properties are
       delete localData[0].evidence; // Evidence is not required and will not throw an error
+      delete localData[0].observationStatus; // Observation Status is not required and will not throw an error
```
Dtphelan1, line 35: DP1 - Could we make sure that this bare-bones object can be fed into the function without throwing errors? Let me know if that makes sense as a test condition before the next step.

Dtphelan1:
With the exception of this very small DP1 observation, I'm good to approve!

test/extractors/CSVCancerDiseaseStatusExtractor.test.js
jafeltra: I added an `expect` statement to check that using the object with all required properties and no optional properties does not throw an error, which I think covers DP1! Also, I believe that the task for updating the CancerDiseaseStatus template will also implement the additional check that the template will render for every possible iteration of optional arguments.

jafeltra:
Rebased and DP1 should be set!

kjmahalingam (approved):
Tested it and it works well!

Dtphelan1:
Ditto - LGTM

Dtphelan1 (approved):

PR #39, opened 2020-10-05
Condition template revamp

julianxcarter:
# Summary
The Condition template has been updated with the new motif pioneered by @Dtphelan1. 
## New behavior
Rather than generating the Condition resource as a json in a `.ejs` file, a JavaScript object is created in a `.js` file by way of reusable functions. Each element of the Condition resource has it's own function.
## Code changes

1. The pre-existing Condition template at `src/templates/Contion.ejs` has been removed and replaced by `ConditionTemplate.js` in the same directory. The new template motif receives each element as an argument in the main module, then uses individual functions to generate objects for each element of the Condition resource. A JavaScript object representing the extracted Condition resource is returned from the main module.

2. References to the `ejs` implementation of the Condition template have been removed in `src/helpers/ejsUtils.js` and `src/templates/index.js`. They've been replaced with references to the `.js` implementation.

3. The unit tests in `test/templates/condition.test.js` have also been updated to test based on the new JavaScript implementation of the Condition template, rather than the `ejs` implementation.

4. Minor update to the example Condition bundle at `test/extractors/fixtures/csv-condition-bundle.json` to include the unique Id generated by the sha256 algorithm.

# Testing guidance
Within `base-care-extraction-client`, make sure that the location of the `mcode-extraction-framework` dependency within `package.json` is set to the local location of the framework. Happy testing :).


src/templates/ConditionTemplate.js
```diff
@@ -0,0 +1,144 @@
...
+    ],
+  };
+}
+
+function categoryArrayTemplate(array) {
+  const category = array.map(individualCategoryTemplate);
+  category.push({
+    coding: [coding({
+      system: 'http://snomed.info/sct',
+      code: '64572001',
```
Dtphelan1, line 63: DP1 - Do we know what this magic code is? If so, would love a comment with explanation

src/templates/ConditionTemplate.js
```diff
@@ -0,0 +1,144 @@
...
+    resourceType: 'Condition',
+    id: conditionId,
+    ...extension(
+      ifAllArgsObj(dateOfDiagnosisTemplate)({ dateOfDiagnosis }),
+      ifAllArgsObj(histologyTemplate)({ histology }),
+    ),
+    ...ifAllArgsObj(clinicalStatusTemplate)({ clinicalStatus }),
+    ...ifAllArgsObj(verificationStatusTemplate)({ verificationStatus }),
+    ...ifAllArgsObj(categoryArrayTemplate)(category),
+    ...ifAllArgsObj(codingTemplate)({ code }),
```
Dtphelan1, line 136: DP2 - If fields are required, we don't need to perform the ifAllArgsObj check. This should hold for the category, code and mrn properties on this object.

test/templates/condition.test.js
```diff
@@ -107,28 +105,20 @@ const CONDITION_INVALID_DATA = {
...
-    const generatedCondition = renderTemplate(
-      CONDITION_TEMPLATE,
-      CONDITION_MINIMAL_DATA,
-    );
+    const generatedCondition = conditionTemplate(CONDITION_MINIMAL_DATA);
 
     expect(generatedCondition).toEqual(minimalValidExampleCondition);
   });
 
   test('invalid data should throw an error', () => {
```
Dtphelan1, line 121: DP3 - One of the new patterns I introduced with the help of JA in the last PR was using a testing utility called `allOptionalKeyCombinationsNotThrow` (verbose, but hopefully clear). I think we could use that here, and mirror the usage in `patient.test.js`

Dtphelan1 (changes requested):

src/extractors/CSVConditionExtractor.js
```diff
@@ -68,7 +65,8 @@ class CSVConditionExtractor extends Extractor {
 
   async getConditionData(mrn) {
     logger.debug('Getting Condition Data');
-    return this.csvModule.get('mrn', mrn);
+    const data = await this.csvModule.get('mrn', mrn);
+    return data[0];
```
Dtphelan1, line 69: DP4 - Curious about the motivation for this change. I worry that this will limit us to only returning a single row of condition data, even if there are multiple to gathe. @julianxcarter help me understand what's happening here 😄

Dtphelan1 (approved):

Dtphelan1 (changes requested):
Disregard my previous approval - sloppy fingers strike again.

src/extractors/CSVConditionExtractor.js
Dtphelan1: 🤦  re: the approval.

src/templates/ConditionTemplate.js
julianxcarter: DP2 - Gotcha, I'll remove the check where it's not needed!

src/templates/ConditionTemplate.js
julianxcarter: DP1 - This is a fixed value for the category element on the Condition resource! The actual value set for this can be found [here](http://hl7.org/fhir/us/core/STU3.1/ValueSet-us-core-condition-category.html), but to my knowledge this code needs to be included.

kjmahalingam (approved):
Beyond DP's comments, I have no additional thoughts on this and it looks good and works well!

src/templates/ConditionTemplate.js
Dtphelan1: This looks good - could we add a comment to the code explaining this for future me who will invariable forget. 😄

src/extractors/CSVConditionExtractor.js
julianxcarter: DP4 - I may be misunderstanding how the extractor is intended to be used. I found that when `csvModule.get()` was called, it only ever returned an array containing a single element; with that element being the current row of the condition data. Because of that I figured it would be simpler to just return that single element, rather than the whole array and then call a `.map()` on a singular element within the formatting function. This may only be the case for our sample data though, so it can be changed back if you have concerns!

src/templates/ConditionTemplate.js
Dtphelan1: Thanks for the fix!

src/extractors/CSVConditionExtractor.js
Dtphelan1: The last part of your thought should be spot on; I think this is a product of the test/sample data, not the extractor proper. Since there is only one condition associated with this MRN in the sample data array, it will only return one element; in the wild, many conditions can be associated with the patient in the `condition.csv`. In fact, I would have hoped that a test would have broken after you made this change... Which brings me to: 

DP4, cont. - I think we should revert this change. 
DP5 - we should probably add a test to the CSVConditionExtractor that tests our ability to extract multiple conditions when there are multiple. This an important piece of functionality that we probably want to ensure works in every batch of the code we deploy. LMK if you think this is overstepping.

julianxcarter: DP 4 & 5: Not an overreach at all, this sounds like a great precaution. I didn't think about the fact that in reality there could be multiple conditions for a single patient. I'll go ahead and revert this code, and also add the new test case. Nice catch!

src/extractors/CSVConditionExtractor.js
```diff
@@ -17,10 +17,8 @@ function formatData(conditionData) {
       throw new Error('The condition is missing an expected attribute. Condition id, mrn, code system, code, and category are all required.');
     }
     return {
-      id: conditionId,
-      subject: {
-        id: mrn,
-      },
+      conditionId,
+      mrn,
```
Dtphelan1, line 21: In this context this change is fine, but keep in mind that the template-data interface can be shared across multiple extractors. Changes to the schema/format of data digested by templates needs to be propagated across any other extractors that use those templates. Obviously in this case we're good, since the EpicCondition extractor doesn't use templates. Just wanted to call this out explicitly 😄

Dtphelan1: .... I spoke too soon

DP4?5? - This breaks an assumption that the `ejsUtils.js` makes with respect to resource id's always being stored under the `id` field. This is why an error was being thrown on the condition-bundle fixture, which I think was fixed by changing the fixture. I think the fixture was right before, and we might just maintain the `id: conditionId` field as it was. Feel free to keep the mrn change. I'll tag the place where this still breaks in the condition-bundle fixture

Dtphelan1:
Needs a rebase, and there is an opportunity to use the snippets Dan has made (`valueCodeableConcept` and `reference`). Happy to talk about details more tomorrow, but hopefully those won't be crazy changes! 😄

Dtphelan1 (approved):

Dtphelan1:
I think this looks good to me!

src/templates/ConditionTemplate.js
```diff
@@ -0,0 +1,115 @@
+const { extension, coding, valueCodeableConcept, reference } = require('./snippets');
+const { ifAllArgsObj } = require('../helpers/templateUtils');
+
+function histologyTemplate({ histology }) {
+  return {
+    url: histology.url,
+    ...valueCodeableConcept({ ...histology }),
```
danlee1025, line 7: I think if you're not passing in any additional properties you can just pass in `histology` without using the spread operator.
```suggestion
    ...valueCodeableConcept(histology),
```

I think there's several other instances with `coding` that this change can be made also in this file.

Dtphelan1: oooh that's a good catch.

src/templates/ConditionTemplate.js
```diff
@@ -0,0 +1,115 @@
...
+}
+
+function subjectTemplate({ subject }) {
+  return {
+    subject: reference(subject),
+  };
+}
+
+// Based on https://mcodeinitiative.github.io/StructureDefinition-obf-Condition.html
+// Official url: http://hl7.org/fhir/us/mcode/StructureDefinition/obf-Condition
```
Dtphelan1, line 89: Dan's catch above made me realize I don't fully know what the shape of those arguments are. Could we have as a comment the schema for the arguments to this Template function. We don't have that for the other templates until now because they've been pretty flat in terms of incoming args. This template, for reasons we can talk about later, still uses complex objects as input arguments and it's not totally clear what the expected shape of those arguments is in the code right now. I've copied and pasted this information as it was in the old .ejs template: 

```
{
  id: String,
  assertedDate: {
    valueDateTime: DateTime
  },
  histologyMorphologyBehavior: {
    system: String,
    code: String,
    cisplay: String
  },
  clinicalStatus: {
    system: String,
    code: String
  },
  verificationStatus: {
    system: String,
    code: String
  },
  category: {
    system: String,
    code: String
  },
  code: {
    system: String,
    code: String,
    display: String
  },
  bodySite: [{
    laterality: {
      system: String,
      code: String
    },
    system: String,
    code: String
  }],
  subject: {
    id: String,
    name: String,
  },
}
```

julianxcarter: Absolutely, this would be a great addition!

Dtphelan1 (changes requested):
to avoid accidental merge, removing approval while small changes are fixed.

src/extractors/CSVConditionExtractor.js
```diff
@@ -26,36 +26,13 @@ function formatData(conditionData) {
...
-        system: 'http://snomed.info/sct',
-        code: laterality,
-        url: 'http://hl7.org/fhir/us/mcode/StructureDefinition/mcode-laterality',
-      },
-      histology: !histology ? null : {
-        system: 'http://snomed.info/sct',
-        code: histology,
-        url: 'http://hl7.org/fhir/us/mcode/StructureDefinition/mcode-histology-morphology-behavior',
-      },
+      category: category.split('|'),
```
kjmahalingam, line 29: DM1: It looks like in flattening the arguments, we might have lost the input of the systems and urls as arguments for the template. This seems like it shouldn't be a huge concern at the surface level, but the difficulty is that we now assume that since we've moved these systems and URLs into the templates, *every* condition we ever want to create, whether using this extractor or not, requires the SNOMED code system, mCODE laterality URL, etc. I think that flattening like you did is definitely a good thing, but maybe we also want to keep all the data discussed. Such as breaking the above `histology` object into `histologySystem`, `histologyCode`, and `histologyUrl`. Does that make sense? If you disagree for any reason, let me know and we can talk about it.

kjmahalingam (changes requested):
This all looks really good! One more nitpicky comment made in the PR. I'm "requesting changes" but that's just to prevent accidental merge. If the answer is that we don't want to address this right now and we're fine with this assumption, works for me.

kjmahalingam (approved):

Dtphelan1 (approved):
I think this looks good! Great work Julian!!!

PR #40, opened 2020-10-05
CDS Template Revamp

danlee1025:
# Summary
Cancer Disease Status template updated with JS snippet approach.
## New behavior & Code Changes
- Removed `CancerDiseaseStatus.ejs` and created a `CancerDiseaseStatusTemplate.js`.
- Added `reference` and `valueCodeableConcept` snippets.
    - I made this separate from the `valueX` snippet since it will be a lot easier to implement once we implement the task for optional args. It seems `valueCoding` and `valueCodeableConcept` are very similar so I didn't want to introduce complex logic to differentiate these two since we'll probably change it with the optional arg approach.
- Updated tests and fixtures.
# Testing guidance
- Ensure tests pass and clients still work with the updated template.


test/templates/snippets/reference.test.js
```diff
@@ -0,0 +1,30 @@
...
+  test('Throws an error when empty object is supplied', () => {
+    expect(() => reference({})).toThrowError(Error);
+  });
+
+  test('Returns minimal reference object when minimal arguments are supplied.', () => {
+    const MINIMAL_DATA = {
+      id: 'example-id',
+    };
+
+    expect(reference(MINIMAL_DATA)).toEqual(minimalReference);
```
Dtphelan1, line 19: DP1 - Small, but could we add a line for the `allOptionalKeyCombinationsNotThrow` function check in this minimal block or in another one as well?

Dtphelan1: If you're not familiar, that's a test-utility function I added in my past PR; you can see it in use in the `patient.test.js` file

danlee1025: yea I was thinking about it but do you think it's necessary where there's only one optional argument? Maybe I don't fully understand what `allOptionalKeyCombinationsNotThrow` does 😂

Dtphelan1: my only thinking is that Reference might get more complicated over time, so it might be good to add it now.

Dtphelan1:
Other than DP1 this has my approval.

jafeltra:
This looks good! The one thing I noticed (mostly because I noticed on my task as well), but when we use our look up functions to go from some display text to a code, we set the codes to just be numbers, but the FHIR resources should have them as strings. I think the EJS templates must have been making them into strings behind the scenes for us, but now we need to handle it ourselves. I think the best way to handle this (but mostly because this is how I did it, so this is very biased), is to just change the three lookup functions in `diseaseStatusUtils` [here](https://github.com/mcode/mcode-extraction-framework/blob/master/src/helpers/diseaseStatusUtils.js#L11) to return the codes as strings instead of numbers.

A picture of the diff I'm getting that I think this should resolve: 
image
(There is a similar diff for the evidence code as well).

danlee1025:
DP1 and JA1 should be addressed.

jafeltra (approved):
This looks good to me! JA1 is resolved!

Dtphelan1 (approved):
Looks pretty good to me!

PR #41, opened 2020-10-06
Care Plan template revamp

jafeltra:
# Summary
Updates the CarePlanWithReview template to use the new templating approach.

## Code changes
- Removed the old `CarePlanWithReview.ejs` and replaced with the `CarePlanWithReviewTemplate.js`
- Added a few snippets for top level properties that are on all [domain resources](https://www.hl7.org/fhir/domainresource.html), since I figured they were used in a few places. They're pretty simple so they also just be functions in the Care Plan template file if that seems more reasonable.
- Updated the CarePlan template tests to test all cases, following those outlined in the Patient template tests.

# Testing guidance
~I think this should probably be merged after #40, which will let me update to use the `reference` and `valueCodeableConcept` snippets Dan added.~ Updated and rebased!

Ensure the tests pass and clients output the same data as they did previously.


src/templates/snippets/resource.js
```diff
@@ -0,0 +1,19 @@
...
+      profile: profiles.map((p) => p),
+    },
+  };
+}
+
+function narrative(status, div) {
+  return {
+    status,
+    div,
+  };
```
Dtphelan1, line 13: DP2 - Are these two mandatory fields for narrative? If not we should use the spread motif and optionally apply them, right?

jafeltra: They are both `1..1` on narrative!

src/templates/snippets/resource.js
```diff
@@ -0,0 +1,19 @@
+function meta(profiles) {
+  return {
+    meta: {
+      profile: profiles.map((p) => p),
```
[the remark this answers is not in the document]
jafeltra: Ah yes this makes sense. I'll udpate!

Dtphelan1:
Jules this needs a rebase what are _you doing!?!???!?!!_

jafeltra:
Dylan this _is_ rebased what are _you doing?!?!_

jafeltra:
More seriously, DP1 should be addressed, I rebased, and updated to use Dan's `reference` and `valueCodeableConcept` snippets. I did have to update the `valueCodeableConcept` on a bit, so I'm happy to discuss that change!

kjmahalingam (approved):
Looks good and works for me!

src/templates/CarePlanWithReviewTemplate.js
```diff
@@ -0,0 +1,113 @@
...
+    ],
+  };
+}
+
+// Treatment Plan Change modeled with CarePlanWithReview Template
+// Uses the ICARE R4 Care Plan profile which is not published yet
+// For reference, ICARE R4 Care Plan profile: http://standardhealthrecord.org/guides/icare/StructureDefinition-icare-care-plan-with-review.html
+function carePlanWithReviewTemplate({
+  id, effectiveDateTime, effectiveDate, treatmentPlanChange, subject,
+}) {
```
Dtphelan1, line 89: DP - Could we keep the input argument schema as a comment here? I've copied them over from the deleted `.ejs` file: 
```
{
  id: String,
  effectiveDateTime: DateTime,
  effectiveDate: Date,
  treatmentPlanChange: {
    hasChanged: Boolean,
    reason: {
      code: String,
      displayText: String,
    }
  },
  subject: {
    id: String,
    name: String,
  },
}
```


(unless by some act of god the input arguments get flattened somehow....)

jafeltra: By some act of god, the input arguments have been flattened. Let me know what you think of the flattening. I tried to follow what the Patient template did for codes/displayTexts.

jafeltra:
An act of _god????_ 998cfb1

Dtphelan1:
Tests are still passing and integration looks good still. Going through the code manually now to see if anything else stand out, but this is looking good.

Dtphelan1 (approved):
I think this looks good to me 😄

kjmahalingam:
Still good to me!

PR #43, opened 2020-10-08
Observation template revamp

kjmahalingam:
# Summary
Updates the Observation template to use the new templating approach.

## Code changes
- Removed the old `Observation.ejs` and replaced with the `ObservationTemplate.js`
- Flattened and changed the structure of the input arguments to the template in `CSVObservationExtractor.js`
- Add more to `observationUtils.js` to determine the type of Observation to process
- Move generic FHIR operations to `fhirUtils.js`
- Update `valueQuantity` check in `valueX.js`
- Update tests and fixtures

# Testing guidance
Ensure the tests pass and that the output of the `base-icare-extraction-client` (on its corresponding PR) succeeds and looks correct.


src/templates/ObservationTemplate.js
```diff
@@ -0,0 +1,128 @@
...
+            {
+              system: 'http://terminology.hl7.org/CodeSystem/observation-category',
+              code: 'laboratory',
+            },
+          ],
+        },
+      ],
+    };
+  }
+  return null;
```
kjmahalingam, line 36: DM1 (me @ myself): Is returning null safe here? I assume that since we are spreading this operation later inside the context of an object, it will just result in no category being rendered, but I don't want there to be a random `null` hanging around so just clarifying.

src/templates/ObservationTemplate.js
```diff
@@ -0,0 +1,128 @@
...
+  return {
+    subject: reference({ id: subjectId }),
+  };
+}
+
+// TODO: We might want to consider a better way to handle the value field
+// if we revisit the valueX inference approach at a later date
+function valueTemplate({ code, valueCode, valueSystem }) {
+  if (!(code && valueCode)) return null;
+
```
Dtphelan1, line 59:
```suggestion
```

src/helpers/fhirUtils.js
```diff
@@ -1,4 +1,34 @@
 const fhirpath = require('fhirpath');
+const { invertObject } = require('./helperUtils');
+
+// Unit codes and display values fo Vital Signs values
+// Code mapping is based on http://hl7.org/fhir/R4/observation-vitalsigns.html
+const quantityCodeToUnitLookup = {
```
jafeltra, line 6: We talked about this lookup a bit at the working session. I think this is trying to map unit `codes` to the `unit` string representation of the codes. I'm not entirely sure if all the entries in this map are accurate or comprehensive. But after looking at the [Quantity](https://www.hl7.org/fhir/datatypes.html#Quantity) in FHIR, it looks like the unit is not required, so I think it might just be better to not manipulate the code at all, and instead just set it to the `code` field. If we do that though, I believe we'll have to set a `system` as well based on the constraint.

src/helpers/fhirUtils.js
```diff
@@ -1,4 +1,34 @@
...
+  cm: 'cm',
+  kg: 'kg',
+  g: 'g',
+};
+
+const quantityTextToCodeLookup = invertObject(quantityCodeToUnitLookup);
+
+function getQuantityUnit(unitCode) {
+  if (!Object.keys(quantityCodeToUnitLookup).includes(unitCode)) {
+    return null;
```
jafeltra, line 24: If we do end up keeping the look up above, I think we should consider just returning the `unitCode` value here. I think it is better to preserve the potentially incorrect value provided rather than remove it. I think that's a pattern we've followed in other places in the framework as well.

src/templates/snippets/valueX.js
```diff
@@ -36,11 +37,11 @@ function valueX(value) {
...
       }
       if (quantRegex.test(value)) {
-        const quantMatch = quantRegex.test(value);
+        const quantMatch = quantRegex.exec(value);
         return {
           valueQuantity: {
-            value: quantMatch[1],
-            ...(quantMatch[6] && { unit: quantMatch[6] }),
+            value: parseFloat(quantMatch[1], 10),
+            ...(quantMatch[6] && { unit: getQuantityUnit(quantMatch[6]) }),
```
jafeltra, line 44: Based on the comment I made above, I think we might want to set this matched "unit" value to the `code` on a quantity (see [Quantity](https://www.hl7.org/fhir/datatypes.html#Quantity)).

src/templates/ObservationTemplate.js
```diff
@@ -0,0 +1,128 @@
...
+  if (isTumorMarker(code)) return valueCodeableConcept({ code: valueCode, system: valueSystem });
+  if (isECOGPerformanceStatus(code) || isKarnofskyPerformanceStatus(code)) return valueX(parseInt(valueCode, 10));
+  return valueX(valueCode); // Vital Sign will be parsed as quantity, others will be parsed as appropriate
```
jafeltra, line 62: I think we decided that we want quantities to default to using ucum (http://unitsofmeasure.org) as the `system` if one is not specified. I think this might be the place to do that since we'll know that a `valueQuantity` was returned here and we have access to the `valueSystem`.

jafeltra (commented):
I'm still reviewing, but this group of comments are all related to the discussion we had about using `code` vs `unit` for valueQuantities, so I wanted to leave them as a group. I'm going to continue reviewing the rest of this now!

test/helpers/observationUtils.test.js
```diff
@@ -12,16 +13,29 @@ describe('observationUtils', () => {
...
-    });
+  test('isTumorMarker should return false when passed a code that does not belong to a Tumor Marker', () => {
+    const code = '12345';
+    expect(isTumorMarker(code)).toEqual(false);
+  });
+  test('isKarnofskyPerformanceStatus should return true when passed the correct code', () => {
+    const code = '89243-0';
+    expect(isKarnofskyPerformanceStatus(code)).toEqual(true);
+  });
+  test('isKarnofskyPerformanceStatus should return false when passed a code that does not belong to a Vital Sign', () => {
```
jafeltra, line 29: I think there's an old reference to "Vital Sign" here

test/helpers/observationUtils.test.js
```diff
@@ -12,16 +13,29 @@ describe('observationUtils', () => {
...
+    const code = '12345';
+    expect(isKarnofskyPerformanceStatus(code)).toEqual(false);
+  });
+  test('isECOGPerformanceStatus should return true when passed the correct code', () => {
+    const code = '89247-1';
+    expect(isECOGPerformanceStatus(code)).toEqual(true);
+  });
+  test('isECOGPerformanceStatus should return false when passed a code that does not belong to a Vital Sign', () => {
```
jafeltra, line 37: Similar to above,  I think there's an old reference to Vital Sign.

src/templates/ObservationTemplate.js
kjmahalingam: We determined that this is safe, so I retract DM1 from myself.

test/templates/observation.test.js
```diff
@@ -1,93 +1,60 @@
...
 
     // Relevant fields should match the valid FHIR
     expect(generatedObservation).toEqual(minimalObservationResource);
   });
 
   test('invalid data should throw an error', () => {
     // ReferenceError will happen when a required field in the template is undefined
-    expect(() => renderTemplate(OBSERVATION_TEMPLATE, INVALID_DATA)).toThrow(ReferenceError);
+    expect(() => observationTemplate(INVALID_DATA)).toThrow(Error);
   });
```
jafeltra, line 59: I think we might want one additional test here that checks that all combinations of the optional fields can be passed into the template without it crashing. Similar to [this patient template test](https://github.com/mcode/mcode-extraction-framework/blob/master/test/templates/patient.test.js#L45).

jafeltra (commented):
This looks really good! I just had a couple additional minor comments. Let me know if you disagree with any of them!

jafeltra (approved):
This looks great! You've resolved all my comments!

Dtphelan1:
Some additional reading of the FHIR materials on [Quantity](https://www.hl7.org/fhir/datatypes.html#Quantity), the [R4 Observation resource content](https://www.hl7.org/fhir/observation.html#resource) and examples of observations - both [Vitalsigns](https://www.hl7.org/fhir/observation-vitalsigns.html) and plain old [R4 Observations](https://www.hl7.org/fhir/observation-examples.html#) make me think that we want to include the `unit` field _as well as_ the  `code` field. Take a look at what I've linked and let me know if you disagree strongly with my reading.

src/helpers/ejsUtils.js
```diff
@@ -1,19 +1,19 @@
 const ejs = require('ejs');
-const fs = require('fs');
-const path = require('path');
 const _ = require('lodash');
 const shajs = require('sha.js');
 const logger = require('./logger');
 
 const {
-  cancerDiseaseStatusTemplate, carePlanWithReviewTemplate, patientTemplate, researchStudyTemplate, researchSubjectTemplate, conditionTemplate,
+  cancerDiseaseStatusTemplate, carePlanWithReviewTemplate, conditionTemplate, observationTemplate, patientTemplate, researchStudyTemplate, researchSubjectTemplate,
```
Dtphelan1, line 7: Nice rearranging 👍

Dtphelan1:
Bar my comment above re: `unit`, this looks good! (Though a fresh look on Monday morning may reveal something my haggard Friday eyes can't see...)

On that note - have a good weekend!

src/templates/snippets/valueX.js
```diff
@@ -36,11 +36,11 @@ function valueX(value) {
...
       }
       if (quantRegex.test(value)) {
-        const quantMatch = quantRegex.test(value);
+        const quantMatch = quantRegex.exec(value);
         return {
           valueQuantity: {
-            value: quantMatch[1],
-            ...(quantMatch[6] && { unit: quantMatch[6] }),
+            value: parseFloat(`${quantMatch[1]}${quantMatch[2] ? quantMatch[2] : ''}`, 10),
+            ...(quantMatch[6] && { code: quantMatch[6], system: 'http://unitsofmeasure.org' }),
```
julianxcarter, line 43: JC 1: This looks good! The only thing that should be added to the return object is a `unit` property that seems to be a required value in the VitalSigns observations. An example can be seen [here](http://hl7.org/fhir/observation-example-head-circumference.json.html) under the `valueQuantity` element. The unit property is essentially just a display value for the `code` property, so we could restore the lookup table that used to be in `observationUtils` and include that here. Happy to chat more about this!

julianxcarter (changes requested):
This looks and works great! Once my single comment is addressed I'll be ready to approve. 👍🏽

src/templates/snippets/valueX.js
julianxcarter: JC 1 (cont.): Looks like @Dtphelan1 had the exact same comment so this is a bit redundant 😅.

kjmahalingam:
Addressing DP1 and JC1 since they're the same:

Thanks for checking on that! I could re-implement the quantity unit lookup that was implemented before, but in going to do that, I noticed this example.

https://www.hl7.org/fhir/observation-example-respiratory-rate.json.html

It looks like there's a "unit" here for `breaths/minute` that isn't in the map for the code we had originally written. I'm wondering where we got that initial list of units and their mappings, because I am concerned that it might not be complete and we may need to add more. Do @Dtphelan1 or @julianxcarter have any thoughts about this? I will re-implement what we have if that's what's needed, but I don't know if we'd be missing information as a result.

kjmahalingam:
@Dtphelan1 @julianxcarter The latest commit should address your concerns.

julianxcarter (approved):
This looks good to me! I've resolved my comment and approved, nice work :).

jafeltra:
This still looks good to me! I like the addition of the warning being logged and returning the unit code if the map doesn't provide a unit string.

src/helpers/fhirUtils.js
```diff
@@ -1,4 +1,31 @@
 const fhirpath = require('fhirpath');
+const logger = require('./logger');
+
+// Unit codes and display values fo Vital Signs values
+// Code mapping is based on http://hl7.org/fhir/R4/observation-vitalsigns.html
+// and inferred from manually referencing display values of codes in
+// value sets utilized by these Vital Signs
```
Dtphelan1, line 7: Nice.

Dtphelan1:
lgtm

Dtphelan1 (approved):

PR #44, opened 2020-10-12
Category on cancer conditions

jafeltra:
# Summary
Only cancer conditions should include the "disease" category coding.

## New behavior
Previously, we were adding the "disease" category coding to all conditions. This value is only fixed on primary and secondary cancer conditions, so it should only be included in the `category` array on those conditions. Now that we have the flexibility in our templates to support this, we are implementing this change.

## Code changes
Use the util functions to determine if a condition is a cancer condition. Only apply the "disease" fixed value on category if it is a cancer condition.

I also reformatted a couple JSON files, which should be contained to b3f8b7b. The changes for the task should be contained to 322e0f1 if it is easier to see the diffs.

# Testing guidance
Run the test and ensure they pass. Ensure the changes to the tests make sense. I did change to example values used in mocks to represent a cancer code, rather than remove the category code to correctly represent a non-cancer condition, but I can swap to the opposite approach if that's preferred.

### DIFF
--- a/src/templates/ConditionTemplate.js
+++ b/src/templates/ConditionTemplate.js
@@ -1,6 +1,6 @@
 const { extensionArr, coding, valueCodeableConcept, reference } = require('./snippets');
 const { ifAllArgsObj } = require('../helpers/templateUtils');
-const { isConditionCodePrimary, isConditionCodeSecondary } = require('../helpers/conditionUtils');
+const { isConditionCodeCancer } = require('../helpers/conditionUtils');
 
 function histologyTemplate({ histology }) {
   return {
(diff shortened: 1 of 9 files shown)


src/templates/ConditionTemplate.js
```diff
@@ -40,16 +41,19 @@ function individualCategoryTemplate(category) {
...
-  // Including the fixed value for the category element at the end of the array
-  category.push({
-    coding: [coding({
-      system: 'http://snomed.info/sct',
-      code: '64572001',
-    }),
-    ],
-  });
+  const codeValue = `${code}`;
+  if (isConditionCodePrimary(codeValue) || isConditionCodeSecondary(codeValue)) {
```
Dtphelan1, line 47: Small DP1 request: Could we just make this check a utility function in the helpers file? I think we use it in a few places at this point, and it might be better if we just defer to the conditionUtils to determine what is and is not a CancerCondition.

jafeltra:
DP1 should be resolved (and has tests that were totally there the whole time).

jafeltra:
And rebased!

Dtphelan1 (approved):

kjmahalingam (approved):
I approve!

PR #46, opened 2020-10-16
Remove EJS

kjmahalingam:
# Summary
Remove all lingering references to EJS and reformat template import/export structure.
## New behavior
None.
# Testing guidance
Run the tests and lint, and run it with the base client!


jafeltra (commented):
This looks good! I think we should be able to remove the `ejs` dependency in package.json as well!

kjmahalingam:
@jafeltra You're right! I removed it in the last commit.

jafeltra (approved):
This looks entirely excellent. See ya later EJS 👋

Dtphelan1 (approved):
looks great - fantastic work Dylan!

PR #48, opened 2020-10-16
Medication extractor

julianxcarter:
# Summary
CancerRelatedMedicationExtractor has been added to the MEF.
## New behavior
The MEF now contains an extractor and template that that will generate CancerRelatedMedication resources.
## Code changes

1. An example medication statement csv file has been added at `docs/cancer-related-medication.csv`.
2. A template (CancerRelatedMedicationTemplate.js) and extractor (CSVCancerRelatedMedicationExtractor) file have been added to their respective folders.
3. The template method for the new template has been added to the fhirTemplateLookup table in `src/helpers/ejsUtils.js`.
4. Tests for the new extractor and template have been created within the test directory. An example module response and example resource bundles have been added as well.
5. Import and export statements have been added to the `index.js` files within the `src` directory.

# Testing guidance
1. Ensure all tests pass and that there are no linting errors.
2. Run from the updated `base-icare-extraction-client` and ensure bundles are being sent to AWS without issue.


src/templates/CancerRelatedMedicationTemplate.js
```diff
@@ -0,0 +1,83 @@
...
+  return {
+    resourceType: 'MedicationStatement',
+    id,
+    meta: {
+      profile: [
+        'http://hl7.org/fhir/us/mcode/StructureDefinition/mcode-cancer-related-medication-statement',
+      ],
+    },
+    ...extensionArr(ifAllArgsObj(treatmentIntentTemplate)({ treatmentIntent })),
+    status: 'active',
```
julianxcarter, line 73: JC 1 (self-comment): The `status` element is required as part of the mcode spec but not included in the csv format that Rob specifies. I've just included 'active' as a placeholder, but I'm positive that's not what we want in the release. We could potentially infer this status from the `startDate` and `endDate` fields which are both required, since a lot of the values in the [status value set](http://hl7.org/fhir/R4/valueset-medication-statement-status.html) seem to have an indirect relationship with when the medication was taken. This definitely feels like a Rob question, but just food for thought.

src/templates/ResourceGenerator.js
```diff
@@ -18,6 +19,7 @@ const fhirTemplateLookup = {
   Patient: patientTemplate,
   ResearchStudy: researchStudyTemplate,
   ResearchSubject: researchSubjectTemplate,
+  CancerRelatedMedication: cancerRelatedMedicationTemplate,
```
kjmahalingam, line 22: DM1: Super nitpicky, but I think maintaining this list will be easier if we alphabetize it like the rest of it is! Could you shift it into alphabetical order?

src/extractors/CSVCancerRelatedMedicationExtractor.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    return {
+      mrn,
+      // Because medicationId is not a required field, we must make it undefined
+      // when it isn't read in from the module rather that making it null. If the
+      // the value is null, then the bundler function will generate the id element on
+      // the resulting fhir resource with a string value of 'null', rather than encoding
+      // an actual id as we expect it to.
+      id: !medicationId ? undefined : medicationId,
```
kjmahalingam, line 28: DM2: Good observation here! Instead of setting something to `undefined` manually, which I think is a little odd, I think maybe replacing this line with what follows might be preferable.
```
...(medicationId && { id: medicationId }),
```

This is in line with some behavior we have elsewhere in the code. Basically what this does is check whether `medicationId` exists, and if it *does* then we spread out the `id` field from this object into the return object. Otherwise if it *doesn't* exist, then this would evaluate so that there's no field in the object to spread, and we wouldn't even have an `id` field defined at all.

kjmahalingam (commented):
This looks great! I have two very nitpicky comments, but otherwise I think this is looking solid and it works well!

I note for other reviewers that you have to add an extractor for this in your config file; I almost forgot!

src/templates/CancerRelatedMedicationTemplate.js
julianxcarter: JC 1 (self-reply): This was resolved per a conversation with Rob during stand-up today. Next commit will simply include a 'status' field on the csv being imported. Commenting just to document the change.

src/templates/ResourceGenerator.js
julianxcarter: DM1: Ahh yes, that makes a lot of sense. The list has been alphabetized in the most recent commit.

src/extractors/CSVCancerRelatedMedicationExtractor.js
julianxcarter: DM2: I didn't realize that you could use conditionals with the spread operator like that! This was a great tip, thank you :).

kjmahalingam (approved):
Nice work!

Dtphelan1:
Adding this as a comment so we don't lose track, but DP1: We should update the CSV_Templates file in `docs/` to align with new column that was added.

src/extractors/CSVCancerRelatedMedicationExtractor.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    return {
+      mrn,
+      // Because medicationId is not a required field, we must make it undefined
+      // when it isn't read in from the module rather that making it null. If the
+      // the value is null, then the bundler function will generate the id element on
+      // the resulting fhir resource with a string value of 'null', rather than encoding
+      // an actual id as we expect it to.
+      ...(medicationId && { id: medicationId }),
```
Dtphelan1, line 28: DP3: Do we need this comment anymore since DM suggested the clever spread trick above?

julianxcarter: DP2: Probably not since this is used elsewhere in the code base, removed this in the most recent commit.

src/extractors/CSVCancerRelatedMedicationExtractor.js
```diff
@@ -0,0 +1,64 @@
...
+
+  return medicationData.map((medication) => {
+    const {
+      mrn, medicationId, code, codeSystem, displayText, startDate, endDate, treatmentReasonCode, treatmentReasonCodeSystem, treatmentReasonDisplayText, treatmentIntent, status,
+    } = medication;
+
+    if (!(mrn && code && codeSystem && startDate && endDate && status)) {
+      throw new Error('The cancer-related medication is missing an expected element; mrn, code, code system, start date, end date, and status are all required values.');
+    }
+
```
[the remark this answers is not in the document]
julianxcarter: DP3: Taken care of in the most recent commits!

Dtphelan1:
This looks good to me. Thanks Julian!!!

Dtphelan1 (approved):

PR #49, opened 2020-10-20
CSV Procedure Extractor

danlee1025:
# Summary
- Added `CSVProcedureExtractor` and is exported from MEF
## New behavior & Code changes
- Added `ProcedureTemplate` and `CSVProcedureExtractor`
- Added tests for new template and extractor
- Moved `bodySiteTemplate` and `lateralityTemplate`  functions from `ObservationTemplate` to snippet so it could be reused.
# Testing guidance
- Ensure tests pass
- Test extractor by running `base-icare-extraction-client`


julianxcarter (approved):
This looks really good to me, and there's nothing that I'd want added or changed so I'm gonna approve. Nice job!

test/extractors/CSVProcedureExtractor.test.js
```diff
@@ -0,0 +1,65 @@
...
+      const data = await csvProcedureExtractor.get({ mrn: MOCK_PATIENT_MRN });
+      expect(data.resourceType).toEqual('Bundle');
+      expect(data.type).toEqual('collection');
+      expect(data.entry).toBeDefined();
+      expect(data.entry.length).toEqual(1);
+      expect(data.entry).toEqual(exampleCSVProcedureBundle.entry);
+    });
+
+    test('should return empty bundle when no data available from module', async () => {
+      csvModuleSpy.mockReturnValue([]);
```
jafeltra, line 57: When I tried using the extractor and a patient didn't have any procedure information in the CSV, the CSV module threw an error saying `[error]: 	CSV Record with provided key 'mrn' and value was not found`. It seems like maybe the behavior we would want would align more with this test, and it should be okay if there are no procedures for a given patient. However, that error is a pretty high level error in the module, so it might be too much to change this behavior right now. Does this align with what you were thinking? Or is there something I'm missing to have the module return an empty array and not error when there are no procedures?

test/templates/procedure.test.js
```diff
@@ -0,0 +1,79 @@
...
+      id: 'example-id',
+      subjectId: 'example-mrn',
+      effectiveDateTime: '2020-01-01',
+      status: 'completed',
+      code: '152198000',
+      system: 'http://snomed.info/sct',
+    };
+
+    const generatedProcedure = procedureTemplate(PROCEDURE_MINIMAL_DATA);
+    expect(generatedProcedure).toEqual(minimalProcedure);
```
jafeltra, line 18: I realized that this test case should be testing the issue in JA1 and in theory this should show that the missing reason fields produces a valid resource without the `reasonCode` element. However, I looked what the data looked like in the test case I ran, and all the missing fields except for `display` were empty strings. Am I setting up my spreadsheet wrong? Should we be expecting those to be `null` or `undefined`? Or should the template be handling empty strings better? Do the other templates get empty strings for missing fields? (Some of these questions I can look into also, but figured I'd ask them to see if you knew too).

jafeltra (commented):
This looks good to me! I had one comment and a question that I added inline. Let me know if I can clarify either.

test/templates/procedure.test.js
danlee1025: ahh so it looks like I need to pass in `null` from the extractor when the optional fields are empty strings since the template only handles `null` and `undefined`. This seems to be how the other extractors are doing it so I will make that change. In the future we may want handle empty strings better in the template?

src/templates/ProcedureTemplate.js
```diff
@@ -0,0 +1,82 @@
...
+    id,
+    status,
+    code: {
+      coding: [
+        coding({ code, system, display }),
+      ],
+    },
+    subject: reference({ id: subjectId }),
+    performedDateTime: effectiveDateTime,
+    ...ifSomeArgsObj(reasonTemplate)({ reasonCode, reasonCodeSystem, reasonDisplayName }),
```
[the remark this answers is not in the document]
danlee1025: JA1 should be addressed.

test/extractors/CSVProcedureExtractor.test.js
danlee1025: To be honest I just copy and pasted this test. I can't really remember but I assume we put this error in the csv module for a reason? Are there some extractors where we want this error to show up if there isn't any data in the csv? I could specifically catch this error in the procedure extractor if we want the behavior to assign more with the test.

test/templates/procedure.test.js
jafeltra: Yeah I agree that we might want to handle empty strings better in the template. It looks like maybe the patient one is handling it differently since that one doesn't do the null checks in the extractor's formatData function. So I'm not sure what's best, but the fix works!

test/extractors/CSVProcedureExtractor.test.js
jafeltra: I also assume we had that error for a reason but I also don't know what the reason was. It seems like not having a procedure shouldn't be an error because I don't think there's any requirement that patients have procedures, so maybe catching the error in the extractor is reasonable. But it's definitely different than we've done in other places, so maybe it's worth discussing.

src/extractors/CSVProcedureExtractor.js
```diff
@@ -0,0 +1,59 @@
...
+      id: procedureId,
+      subjectId: mrn,
+      status,
+      code,
+      system: codeSystem,
+      display: displayName || null,
+      reasonCode: reasonCode || null,
+      reasonCodeSystem: reasonCodeSystem || null,
+      reasonDisplayName: reasonDisplayName || null,
+      conditionId,
```
jafeltra, line 29: The changes here work! I think that `conditionId` is also optional so we might want it on that as well.

danlee1025: The template actually handles this one haha since it doesn't use the `ifSomeArgsObj`, should I make the change anyway to remain consistent?

jafeltra: Ohh. That makes sense! I say it's fine how it is.

test/extractors/CSVProcedureExtractor.test.js
jafeltra: We've discussed and I added a comment to [this Jira issue](https://jira.mitre.org/browse/STEAM-385) to consider this at a broader scale later.

test/templates/procedure.test.js
jafeltra: We discussed offline and it looks like this is how we handle this in the other extractors, so we'll stick with this approach!

jafeltra (approved):
This looks good to me! Thanks for making that change!

PR #50, opened 2020-10-20
Period hotfix

julianxcarter:
# Summary
Properly handles the omission of the `startDate` and `endDate` fields from a csv. Prior to this, the resource would be generated but the message would fail to send.
## New behavior
`startDate` and `endDate` being omitted from the csv now results in the corresponding fields being left out of the effectivePeriod element, rather than nulling out their respective properties. 
## Code changes
1. Added two conditionals to the `periodTemplate` function within the CancerRelatedMedication template. These conditionals omit either `start` or `end` from the `period` property if they are not included in the csv.
2. Updated the minimal cancer related medication json to reflect the new behavior.
# Testing guidance
When running from the base icare client, remove the startDate and endDate columns within `cancer-related-medication-information.csv` and ensure that the messages are sent to the AWS server.

### DIFF
--- a/src/templates/CancerRelatedMedicationTemplate.js
+++ b/src/templates/CancerRelatedMedicationTemplate.js
@@ -26,8 +26,8 @@ function subjectTemplate({ id }) {
 function periodTemplate({ startDate, endDate }) {
   return {
     effectivePeriod: {
-      start: startDate,
-      end: endDate,
+      ...(startDate && { start: startDate }),
+      ...(endDate && { end: endDate }),
     },
   };
 }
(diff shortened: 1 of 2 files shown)


Dtphelan1:
If things post this is fine, but it feels a bit dirty to have an EffectivePeriod with no information on it.... 😦 . Maybe we want to make another task to ensure that this is a sufficient way of handling this scenario? Will test that this works tomorrow AM, but wanted to jot down my initial reaction before signing off.

Dtphelan1:
The other thought I have is that if this isn't something we caught with our previous tests, we should probably reexamine how we're writing our tests in the minimal-data case; feels like there must be a disconnect between the expectations we have about what a "minimal" case looks like and what it actually looks like.

kjmahalingam (approved):

Dtphelan1 (approved):

PR #51, opened 2020-10-23
ValueX type parameter

julianxcarter:
# Summary
An optional `type` parameter has been added to the `valueX()` snippet. The `valueCodeableConcept()` snippet has also been removed and integrated with the `valueX` snippet.
## New behavior
The `valueX()` function will now default to using the `type` parameter to determine the return type when this parameter is included. If the type parameter is not included; or is included and the specified type is not recognized, the old inferential approach is used. Also, the separate `valueCodeableConcept` snippet has been removed in favor of including `'valueCodeableConcept'` as a parameter in the base `valueX` snippet.
## Code changes
1. `type` was added as an optional parameter to the `valueX()` snippet. A `valueByType()` helper function has also been introduced, returning the appropriate value[x] element based on the type parameter rather than the value parameter. This new helper function returns null if the type parameter is recognizable. 
2. `'valueCodeableConcept'` has been added as a supported type to the `valueX()` snippet. Consequently, the `valueCodeableConcept()` snippet and all of it's references have been removed throughout the codebase and replaced with `valueX()` references with a `'valueCodeableConcept'` argument.
# Testing guidance
Ensure that all tests pass from MEF, also ensure that all bundles still validate when run from the base icare extraction client. Happy Testing!

### DIFF
--- a/src/templates/snippets/valueX.js
+++ b/src/templates/snippets/valueX.js
@@ -8,7 +8,105 @@ const dateRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|
 const dateTimeRegex = /^\d\d\d\d-\d\d-\d\dT\d\d(:\d\d(:\d\d(-\d\d:\d\d)?)?)?$/;
 const valueTimeRegex = /^([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\.[0-9]+)?$/;
 
-function valueX(value) {
+function valueByType(value, type) {
+  switch (type) {
+    case 'valueCodeableConcept':
+      return {
+        valueCodeableConcept: {
+          coding: [
+            coding(value),
+          ],
+          ...(value.text && { text: value.text }),
+        },
+      };
+
+    case 'valueInteger':
+      return {
+        valueInteger: parseInt(value, 10),
+      };
+
+    case 'valueBoolean':
+      return {
+        valueBoolean: value,
+      };
+
+    case 'valueDate':
+      return {
+        valueDate: value,
+      };
+
+    case 'valueDateTime':
+      return {
+        valueDateTime: value,
+      };
+
+    case 'valueTime':
+      return {
+        valueTime: value,
+      };
+
+    case 'valueQuantity':
+      if (typeof value === 'string') {
+        const quantMatch = quantRegex.exec(value);
+        return {
+          valueQuantity: {
+            value: parseFloat(`${quantMatch[1]}${quantMatch[2] ? quantMatch[2] : ''}`, 10),
+            ...(quantMatch[6] && {
+              code: quantMatch[6],
+              system: 'http://unitsofmeasure.org',
+              unit: getQuantityUnit(quantMatch[6]),
+            }),
+          },
+        };
+      }
+      return {
+        valueQuantity: {
+          unit: value.unit,
+          code: value.code,
+          value: value.value,
+          system: 'http://unitsofmeasure.org',
+        },
+      };
+
+    case 'valueString':
+      return {
+        valueString: value,
+      };
+
+    case 'valueRange':
+      return {
+        valueRange: {
+          high: value.high,
+          low: value.low,
+        },
+      };
+
+    case 'valuePeriod':
+      return {
+        valuePeriod: {
+          start: value.start,
+          end: value.end,
+        },
+      };
+
+    case 'valueCoding':
+      return {
+        valueCoding: coding(value),
+      };
+
+    default:
+      logger.warn(`The type ${type} is not recognized by the ValueTemplate, attempting to infer the type of the value.`);
+      return null;
+  }
+}
+
+function valueX(value, type = null) {
+  if (type) {
+    const returnValue = valueByType(value, type);
+    if (returnValue != null) {
+      return returnValue;
+    }
+  }
   switch (typeof value) {
     case 'number':
       return {
@@ -89,19 +187,6 @@ function valueX(value) {
   }
 }
 
-// TODO: move this into valueX when we define optional args
-function valueCodeableConcept(value) {
-  return {
-    valueCodeableConcept: {
-      coding: [
-        coding(value),
-      ],
-      ...(value.text && { text: value.text }),
-    },
-  };
-}
-
 module.exports = {
-  valueCodeableConcept,
   valueX,
 };
--- a/src/templates/snippets/valueX.js
+++ b/src/templates/snippets/valueX.js
@@ -58,14 +58,13 @@ function valueX(value, type = null) {
           valueType = 'valueQuantity';
         } else if (value.system || value.version || value.code || value.display || value.userSelected) {
           valueType = 'valueCoding';
+        } else {
+          logger.debug(`The provided object has an unknown shape, including properties ${Object.keys(value)}`);
         }
         break;
 
       default:
         logger.warn(`Unable to determine the value[x] specialization with the provided value of type - ${typeof value}`);
-        if (typeof value === 'object') {
-          logger.debug(`The provided object has an unknown shape, including properties ${Object.keys(value)}`);
-        }
         break;
     }
   }
--- a/src/templates/snippets/valueX.js
+++ b/src/templates/snippets/valueX.js
@@ -63,6 +63,9 @@ function valueX(value, type = null) {
 
       default:
         logger.warn(`Unable to determine the value[x] specialization with the provided value of type - ${typeof value}`);
+        if (typeof value === 'object') {
+          logger.debug(`The provided object has an unknown shape, including properties ${Object.keys(value)}`);
+        }
         break;
     }
   }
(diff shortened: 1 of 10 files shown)


src/templates/snippets/valueX.js
```diff
@@ -8,7 +8,105 @@ const dateRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|
...
+          ],
+          ...(value.text && { text: value.text }),
+        },
+      };
+
+    case 'valueInteger':
+      return {
+        valueInteger: parseInt(value, 10),
+      };
+
```
kjmahalingam, line 27: DM1: Any chance that we could also add the following case?

```
    case 'valueDecimal':
      return {
        valueInteger: parseFloat(value),
      };
```

It shouldn't be relevant quite yet, but it's a known case that we will likely encounter in the future.

src/templates/snippets/valueX.js
```diff
@@ -8,7 +8,105 @@ const dateRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|
...
+        },
+      };
+
+    case 'valueCoding':
+      return {
+        valueCoding: coding(value),
+      };
+
+    default:
+      logger.warn(`The type ${type} is not recognized by the ValueTemplate, attempting to infer the type of the value.`);
```
kjmahalingam, line 98: DM2: I'm not sure I like having a `warn` level log here. I think that this is a common enough occurrence that it will happen pretty frequently, and I think this isn't particularly relevant info for a user to see. This might be helpful to have at `debug` level, so that we can have the info for our purposes if things go wrong.

kjmahalingam (commented):
This looks good! I left a couple minor comments.

DM3: I don't want this to become too much work, but I wonder...do we want to have a test case for each explicit input type of the function now? It feels like it might be good for the sake of a sanity check, but feel free to push back.

src/templates/snippets/valueX.js
julianxcarter: DM2: Good point, this is definitely developer-level context that an end user wouldn't need to see when running it. I'll downgrade the log in another commit.

jafeltra (commented):
This looks good! I had one minor comment inline. However, I'm also wondering if there is a way to simplify the `valueByType` and `valueX` functions. I noticed there are a few block that are reused, like the block for a `valueQuantity` when it is a string and the block when `valueQuantity` is an object. I noticed that at least that second case, `valueQuantity` when it is an object, is different in the two functions. Is there a reason for that? If not, I'm worried we might have similar cases that would happen in the future and I'm wondering if there's a better way to reuse the different values we want to return. That might require some refactoring and I'm not 100% how it should work or if it would be worth it, so I'm curious to see what you think, and I'm happy to discuss an approach.

src/templates/snippets/valueX.js
```diff
@@ -8,7 +8,110 @@ const dateRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|
...
+      };
+
+    case 'valueInteger':
+      return {
+        valueInteger: parseInt(value, 10),
+      };
+
+    case 'valueDecimal':
+      return {
+        valueInteger: parseFloat(value),
```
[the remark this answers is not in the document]
julianxcarter: Ja1: Ahh yes, fixing this now!

julianxcarter:
> This looks good! I had one minor comment inline. However, I'm also wondering if there is a way to simplify the `valueByType` and `valueX` functions. I noticed there are a few block that are reused, like the block for a `valueQuantity` when it is a string and the block when `valueQuantity` is an object. I noticed that at least that second case, `valueQuantity` when it is an object, is different in the two functions. Is there a reason for that? If not, I'm worried we might have similar cases that would happen in the future and I'm wondering if there's a better way to reuse the different values we want to return. That might require some refactoring and I'm not 100% how it should work or if it would be worth it, so I'm curious to see what you think, and I'm happy to discuss an approach.

So there's actually no good reason for `valueQuantity` being different in the two functions, I had just added the `getQuantityUnit()` call in the `valueX()` function but forgot to do the same in the `valueByType()` function. This, I think, is indicative of a problem. The good news is that there's definitely a way to refactor this so that two functions aren't needed at all so I'll go ahead and consolidate them!

src/templates/snippets/valueX.js
```diff
@@ -8,35 +8,107 @@ const dateRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|
...
+      case 'object':
+        if (value.high || value.low) {
+          valueType = 'valueRange';
+        } else if (value.start || value.end) {
+          valueType = 'valuePeriod';
+        } else if (value.value && value.code) {
+          valueType = 'valueQuantity';
+        } else if (value.system || value.version || value.code || value.display || value.userSelected) {
+          valueType = 'valueCoding';
+        }
```
jafeltra, line 61: JA2: I realized we had a `logger.warn` statement if we reached an unknown object shape in the previous implementation. Do you think we should add a similar warning here?

For reference, this is the old warning:
```
logger.warn(`Trying to render a ValueTemplate with an object of unknown shape, including properties ${Object.keys(value)}`);
```

src/templates/snippets/valueX.js
```diff
@@ -8,35 +8,107 @@ const dateRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|
...
+          valueType = 'valueString';
+        }
+        break;
+
+      case 'object':
+        if (value.high || value.low) {
+          valueType = 'valueRange';
+        } else if (value.start || value.end) {
+          valueType = 'valuePeriod';
+        } else if (value.value && value.code) {
```
jafeltra, line 57: JA3: This is really just a question, but I noticed the previous implementation was checking if there was a `value.unit` when deciding if `valueQuantity` should be inferred based on the object shape. Is there a reason you decided to change it? Maybe it is better this way! I'm just curious.

test/templates/snippets/valueX.test.js
```diff
@@ -0,0 +1,235 @@
...
+  });
+
+  test('Should generate valueDateTime when type parameter is used', () => {
+    const exampleDateTime = '2020-02-01';
+    const exampleValueDateTime = {
+      valueDateTime: '2020-02-01',
+    };
+    expect(valueX(exampleDateTime, 'valueDateTime')).toEqual(exampleValueDateTime);
+  });
+
```
jafeltra, line 83: JA4: I think you should be able to add a test here to check that a `valueDateType` is generated when the parameter is not used and the dateTime has a date and a time, like `2020-02-01T10:35:59`.

jafeltra (commented):
This looks really good! I really like the approach for the refactor! I think it'll be a lot easier to maintain the different cases going forward now! I just had two small comments and a question below.

src/templates/snippets/valueX.js
kjmahalingam: Sorry I steered you wrong with the integer typo in here earlier, thanks for fixing it. :)

kjmahalingam (approved):
Excellent! I agree with Julia's suggested new changes, but aside from that, this is good by my review. :)

kjmahalingam:
This will also need a rebase, since the validation tests have been merged in!

test/templates/snippets/valueX.test.js
julianxcarter: Ja4: You're right, that'd be useful here! Coming soon to a commit near you!

src/templates/snippets/valueX.js
julianxcarter: JA3: There is a reason! I figured that for the most part (given that it's been the case for all of the extractors we've built so far), we would not be reading in a `unit` property from the csv input and would infer them from the `code` property instead. Therefore, I thought the best place for that inference would be in this module as the `valueQuantity` object is being generated. `Quantity`-based types are also the only type with both a `code` and `value` property, so I thought they would be just as effective to use in a conditional for recognizing the object. I'm open to chatting more about this or altering my assumptions!

src/templates/snippets/valueX.js
julianxcarter: JA2: I felt that the `logger.warn` on line 65 was sufficient for that case (shown below).
```
logger.warn(`Unable to determine the value[x] specialization with the provided value of type - ${typeof value}`);
```
Maybe if the provided value is an object, then we could add a `debug` level warning which includes the Object keys like the old warning used to? This seems to lend itself more to a `debug` level warning rather than a `warn` level warning in my mind, but I could swing either way!

src/templates/snippets/valueX.js
jafeltra: Oh nice! That makes sense to me!

src/templates/snippets/valueX.js
jafeltra: I like the debug statement! I think that makes sense!

jafeltra (approved):
This looks good to me! Thanks for making those changes!

PR #52, opened 2020-10-26
Validation tests

kjmahalingam:
# Summary

Adds FHIR JSON Schema validation to the tests for templates, to ensure that all templates are returning valid FHIR.

## New behavior

Adds additional testing, no new functionality otherwise.

## Code changes

Adds a new util function called `isValidFHIR` that is used across all template tests. Also fixes a few fixtures and test data that were invalid according to FHIR.

# Testing guidance

Run tests and lint and make sure they work. I noticed on my end that some of the tests are slow because of the JSON Schema validation being done, but I have also noticed that my machine's jest has been slow in general. If others notice the tests are particularly slow, we should talk about that.


julianxcarter (approved):
This looks great to me! Also, on my laptop all of the unit tests only take between 11 and 13 seconds to run which I think is still pretty good. I've verified that the json validation is still happening here, so I don't think that speed will be an issue moving forward with this. Great work!

jafeltra (approved):
This makes sense to me! When I ran the tests it took between 11 and 17 seconds, but it never felt crazy slow or anything to me, so I'm good with this! The actual test makes sense to me!

PR #53, opened 2020-10-27
Improve logging when OperationOutcomes are present

jafeltra:
# Summary
When an `OperationOutcome` is present, we now log more detailed information from that resource.

## New behavior
Right now, I think the only time we could get an `OperationOutcome` is from a `search` from the BaseFHIRModule. When we get the bundle back from the search, we should look to see if there is an [OperationOutcome](https://www.hl7.org/fhir/operationoutcome.html) resource on it. If there is, we log the information on that resource. Severity and code are required on the resource, so I assume those are always present. If other information is present, it is added. I thought the list of codes was a lot of information in the log statements, so I put that list in a `debug` so it is only logged if we use `--debug`.

## Code changes
Added a helper function to do the logging for OperationOutcome information.

# Testing guidance
Test to make sure the information logged is helpful and formatted in an appropriate way. (It is basically a long sentence right now based on a brief discussion last week.) In order to get get search results with an OperationOutcome when using the ICARE Client, you can either include the Procedure Extractor in your config file or include the MedicationRequest Extractor and change the default status to an invalid value [here](https://github.com/mcode/mcode-extraction-framework/blob/master/src/extractors/FHIRMedicationRequestExtractor.js#L3).

I had difficulty writing a test for the new addition to the `search` in BaseFHIRModule. I had tried to spy on `logOperationOutcomeInfo` to see if it was called but was never able to spy on it correctly. I've added a note to the test in case we learn how to do this in the future. I also added a task to the backlog to consider checking the statements that get logged, which could be another way to test the function in the future.

### DIFF
--- a/src/extractors/BaseFHIRExtractor.js
+++ b/src/extractors/BaseFHIRExtractor.js
@@ -1,6 +1,6 @@
 const { Extractor } = require('./Extractor');
 const { BaseFHIRModule } = require('../modules');
-const { determineVersion, getBundleResourcesByType, isBundleEmpty, mapFHIRVersions } = require('../helpers/fhirUtils');
+const { determineVersion, mapFHIRVersions, isBundleEmpty, getBundleResourcesByType } = require('../helpers/fhirUtils');
 const logger = require('../helpers/logger');
 
 function parseContextForPatientId(context) {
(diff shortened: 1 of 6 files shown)


kjmahalingam (approved):
Looks good to me! Nice work!

jafeltra:
@kjmahalingam and @julianxcarter I moved an initial log statement about the presence of an OperationOutcome to the top of the logging function. I had initial thought to use the same severity as the issues for that message, but I realized the issues could all have different severities, so I decided to make the initial message for `An OperationOutcome was returned with the following issue(s):` just be an info message. Let me know what you think of this approach!

kjmahalingam:
I still approve!

julianxcarter (approved):
This looks good to me! Approving this, great job! 👍🏽

PR #56, opened 2020-11-16
Data Absent Reason extension

jafeltra:
# Summary

If both start and end date are not provided in the related medication CSV, the data absent reason extension is used in the `effectivePeriod` element.

## New behavior
The extension will be used in the `effectivePeriod` element if there is no start and no end date provided in the CSV. I decided to use "unknown" as the reason code because the description said "The value is expected to exist but is not known." which I thought was most applicable to the situation where we are missing dates.

## Code changes
I made the extension a snippet since I figured there may be other places we'd want to use it in the future. You can provide the specific reason code when using the snippet.

# Testing guidance
I tested using the VS Code FHIR Validator extension, but ensure that the output now validates (at least around the MedicationStatement effectivePeriod). Ensure tests and code make sense.

### DIFF
--- a/src/templates/CancerRelatedMedicationTemplate.js
+++ b/src/templates/CancerRelatedMedicationTemplate.js
@@ -1,4 +1,10 @@
-const { coding, extensionArr, reference, valueX } = require('./snippets');
+const {
+  coding,
+  dataAbsentReasonExtension,
+  extensionArr,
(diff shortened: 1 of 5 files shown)


kjmahalingam (approved):
Great work!

danlee1025 (approved):

PR #57, opened 2020-11-16
Validate against FHIR in BaseClient after extraction is complete

kjmahalingam:
# Summary

This adds functionality to validate the extracted Bundle in the BaseClient against FHIR after extraction is complete. Since the BaseClient is inherited by all other clients used, this should be the only place this change is necessary. This is useful because now that we have the standalone mCODE client here, we don't have validation on the server side and will want it here instead.

## New behavior

Logs an error on extraction if the Bundle fails FHIR validation.

## Code changes

- Moves the isValidFHIR operation to the fhirUtils and out of test fixtures, and updates the package and path accordingly.
- Adds validation and error logging to BaseClient.

# Testing guidance

Run the tests and lint. Also run the `mcode-extraction-framework` CLI in here, both with valid data to show that no error is logged, and then with invalid data to show that the error is logged.


src/client/BaseClient.js
```diff
@@ -69,6 +70,10 @@ class BaseClient {
       }
     }, Promise.resolve(contextBundle));
 
+    if (!isValidFHIR(contextBundle)) {
+      logger.error('Extracted bundle is not valid FHIR.');
```
kjmahalingam, line 74: I think that we want to just log in here. If we decide we want to halt the process here, that's also fine, but I think that is very aggressive and lean toward just notifying the user.

jafeltra: I think this seems like a reasonable decision! Similar to how the ICARE infrastructure works, whoever uses the bundle can decide if invalid FHIR is allowed or not.

jafeltra (approved):
This makes sense to me! I tested out running the client in MEF and in base-icare-client and both worked as expected. Looks great!

danlee1025 (approved):
LGTM

PR #58, opened 2020-11-18
Standardized empty data approach

julianxcarter:
# Summary
Standardizes the shape of empty data that's fed into the templates.  
## New behavior
A helper function has been introduced to `ResourceGenerator.js` which iterates through object properties and converts all `undefined` and empty string values to `null`.
## Code changes
* A `cleanEmptyData()` method has been added to `ResourceGenerator.js`.
* Unit tests for this method have been added to `resourceGenerator.test.js`.
* Manual empty data setting within each of the extractors has been removed, as they're no longer needed (unless the property in question is being formatted by a separate function such as `formatDateTime()`.
# Testing guidance
Ensure all tests pass, I also suggest manually setting a property to `undefined` or an empty string within one the `formatData()` method of one of the extractors, and then ensuring that the same property comes in to the corresponding template as a `null` value. Happy testing!


src/templates/ResourceGenerator.js
```diff
@@ -37,6 +37,24 @@ function generateResourceId(data) {
...
 
+// Ensures that empty data in the resource object carries a null value, rather than being undefined or an empty string
+function cleanEmptyData(data) {
+  const cleanData = data;
+  Object.keys(cleanData).forEach((element) => {
+    if (typeof cleanData[element] === 'object' && cleanData[element]) {
+      Object.keys(cleanData[element]).forEach((key) => {
+        if (cleanData[element][key] === '' || cleanData[element][key] === undefined) {
+          cleanData[element][key] = null;
+        }
```
Dtphelan1, line 48: DP1 - it looks like this only goes 2 layers deep in this check; should we have a recursive loop here?

julianxcarter: DP1- That'd absolutely be the safest and most inclusive to do here, I'll go ahead and make this recursive.

jafeltra:
This looks really good! I like this approach! Besides Dylan's one comment, the only other thought I had was about using this approach in the template tests. I'm not entirely sure if this is an approach we want to take, but I wanted to throw it out there to see what you and others think. Do you think it makes sense to set up the "minimal" data objects we create with the empty string/null/undefined values and use the `cleanEmptyData` function on those objects? I do think that your tests for the `cleanEmptyData` function test out the functionality appropriately, so I don't really know that this is necessary at all.

src/templates/ResourceGenerator.js
```diff
@@ -37,6 +37,20 @@ function generateResourceId(data) {
...
+    if (typeof cleanData[element] === 'object' && cleanData[element]) {
+      cleanEmptyData(cleanData[element]);
+    }
+    if (cleanData[element] === '' || cleanData[element] === undefined) {
+      cleanData[element] = null;
+    }
+  });
+  return cleanData;
+}
+
```
Dtphelan1, line 53: Alright one last thing on the loop - JavaScript objects can contain circular references and infinite loops. Could we add a MAX_DEPTH parameter to this file, and make sure that we never go beyond a certain maximum depth size? We can then set the default depth parameter to 0, and make sure we increment the depth every time we make a recursive call.

jafeltra:
Okay after thinking about my comment again and talking to @julianxcarter a bit about it, I don't think we'll need to actually use the `cleanEmptyData` function in the template tests. But maybe we should set up the data that we pass into the templates in the tests in a way that assumes it's gone through the `cleanEmptyData` function already. I think the majority of the template tests already do this, but I at least noticed that the care plan template just leaves off the non-required data in the minimal data objects. I think we'd maybe want to have all the non-required data have a `null` value now, right? Git blame seems to be blaming me on this one, so if it helps, I can help out with this.

src/templates/ResourceGenerator.js
julianxcarter: I've added this to the most recent commit. I thought that 3 recursive calls was a good ceiling, but I can raise or lower that depending on what you think is appropriate!

src/templates/ResourceGenerator.js
```diff
@@ -37,6 +37,28 @@ function generateResourceId(data) {
   return shajs('sha256').update(JSON.stringify(data)).digest('hex');
 }
 
+// Ensures that empty data in the resource object carries a null value, rather than being undefined or an empty string
+function cleanEmptyData(data, depth = 0) {
+  const cleanData = data;
+  const MAX_DEPTH = 3;
+  Object.keys(cleanData).forEach((element) => {
+    if (typeof cleanData[element] === 'object' && cleanData[element]) {
+      if (depth < MAX_DEPTH) cleanEmptyData(cleanData[element], depth + 1);
```
Dtphelan1, line 46: ..... okay one last thing, and I mean it this time.

1) Could we make the max_depth something a little more generous, like 50? Hopefully this safeguard should only prevent infinite loops, and not stop us from traversing deep objects.
2) Could we log a warning message when we hit our max depth? Something indicating that we're trying to clean an empty object, hit a maximum recursive depth, and might not totally clean the object of empty values? 
3) Could we write a test case that ensures we halt execution when depths get too large? We could test that a nested object doesn't get changed when the depth is higher than the limit, and that it does when it's smaller. 

I promise I'll be more explicit next time I plan on being this needy 😄

jafeltra (commented):
This looks good to me! I think your latest changes to the template tests make a lot of sense and will help us test if the templates are ever not working as we expect after we use the `cleanEmptyData` function! I didn't see anything else besides Dylan's last few comments, so I'm good to approve ones those are set!

src/templates/ResourceGenerator.js
julianxcarter: Yes to all three! I had a hunch that a limit of 3 may have been outrageously small so this isn't exactly a surprise. I'll push up these changes in the next commit :)

Dtphelan1:
I feel pretty good about this!

Dtphelan1 (approved):

Dtphelan1 (changes requested):
Alright. I'm no longer making promises about the future state of nitpickiness....

Here are some more comments I have after talking with Julia: 
1. We should write a test that explicitly checks that we can handle circular dependencies without failing or blowing the stack. Something like this?
```
  test('Handles circular references without crashing', () => {
    const obj = {};
    obj.obj = obj;
    expect(() => cleanEmptyData(obj)).not.toThrow();
  });
```
2. Instead of building a super-deep object, we could make a shallow one and just manually provide a `depth` argument that is >= than the MAX_DEPTH. Could be 50, 200, your choice. :smile: 
3. In our emptyData else loop, we have an additional clean step. I think we could probably get rid of that, right? When Jules and I were testing this function manually, we were kindof shocked that it cleaned up keys at layers deeper than MAX_DEPTH. 
4. When we loop over the data object's keys, we call each item we loop over an `element`. Could we rename that to `key`, like we do in our secondary loop around line 47/48?

jafeltra (approved):
This looks good to me! The latest changes address the last set of comments and I haven't noticed anything else!

Dtphelan1 (approved):
Thank you so much for your patience. This all looks very good to me 😄

PR #61, opened 2020-12-01
antlr resolution added and preinstall script added

Dtphelan1:
Deals with a weird bug noticed by Julian in his previous PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -791,9 +791,7 @@
       }
     },
     "antlr4": {
-      "version": "4.8.0",
-      "resolved": "https://registry.npmjs.org/antlr4/-/antlr4-4.8.0.tgz",
-      "integrity": "sha512-en/MxQ4OkPgGJQ3wD/muzj1uDnFSzdFIhc2+c6bHZokWkuBb6RRvFjpWhPxWLbgQvaEzldJZ0GSQpfSAaE3hqg=="
+      "version": "4.8.0"
     },
     "anymatch": {
       "version": "2.0.0",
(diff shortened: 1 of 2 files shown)


kjmahalingam:
I don't think I understand what the issue here was (or what antlr is actually even needed for in our code...). Can @Dtphelan1 or @julianxcarter explain this?

julianxcarter:
> I don't think I understand what the issue here was (or what antlr is actually even needed for in our code...). Can @Dtphelan1 or @julianxcarter explain this?

Yes! `fhirpath` is a dependency of the MEF, and `fhirpath` also has `antlr4` as a dependency. When you run `npm upgrade mcode-extraction-framework` from some (but apparently not all) of the repos that rely on the MEF, `antlr4` is upgraded as well from version `4.8.0` to version `4.9.0` which breaks our unit tests. This PR is meant to lock `antlr4` at version `4.8.0` until we can figure out what exactly is going on with this new `antlr4` version, so this issue won't pop up again if we ever need to update the MEF commit from one of our other repos that have it as a dependency. @kjmahalingam

Dtphelan1:
Some more info behind the scenes: 

`antlr` recently upgraded their version from 4.8 to 4.9. `antlr` is a dependency of `fhirpath`. `fhirpath` doesn't specify a locked version number for `antlr` in their `package.json`; instead, they allow any minor/patch updates from 4.7 upward (i.e. `^4.7.0`)

Problem: `antlr` doesn't obey semantic versioning, and version 4.9 requires a new version of node while older versions did not. 

To avoid upgrading issues with other packages that depend on the MEF (see Julian's comment above) we're ensuring that the version `antlr` we use for fhirpath resolves to 4.8.

kjmahalingam (approved):
Consider me convinced!

julianxcarter (approved):

PR #63, opened 2020-12-02
Staging method in the form of a `stagingSystem` property 

Dtphelan1:
# Summary
Adds an `stagingSystem` column to staging CSV schema, where rows can supply an optional stagingSystem code. Adds a new snippet template, `stagingMethodTemplate`, that gets used across both the Staging and TNMCategory templates to translate information about a `stagingSystem` into a `Observation.method` property. Also added a new ValueSet corresponding to the valid StagingSystem codes as a part of this. 

## New behavior
See above

## Code changes
- Updates the spreadsheet and the example staging csv to align with this new format;
- Exposes this new `stagingSystem` property at the CSVStagingExtractor level;
- Updates the Staging and TNMCategory templates to use this new property.
- Adds a new `stagingMethodTemplate` snippet that is used by the two templates above
- Tests this new snippet 
- Creates a stagingUtils file to house Staging-related methods and lookups
- Update existing tests and fixtures based on new data-expectations

# Testing guidance
Ensure tests pass and that the new tests are sufficient for covering this new template


test/sample-client-data/staging-information.csv
```diff
@@ -1,6 +1,6 @@
...
-123,conditionId-1,3C,cT3,cN3,cM0,Clinical,2020-01-01
-456,conditionId-2,3C,cT3,cN3,cM0,Clinical,2020-01-01
-789,conditionId-2,3C,,cN3,cM0,Clinical,2020-01-02
-789,conditionId-2,3C,cT3,,cM0,Clinical,2020-01-03
-789,conditionId-2,3C,cT3,cN3,,Clinical,2020-01-04
+mrn,conditionId,stageGroup,t,n,m,type,stagingSystem,effectiveDate
+123,conditionId-1,3C,cT3,cN3,cM0,Clinical,C146985,2020-01-01
+456,conditionId-2,3C,cT3,cN3,cM0,Clinical,443830009,2020-01-01
+789,conditionId-2,3C,,cN3,cM0,Clinical,444256004,2020-01-02
+789,conditionId-2,3C,cT3,,cM0,Clinical,443830009,2020-01-03
```
jafeltra, line 5: I don't think this really matters at all, but do you think it would be useful to have one of these rows (maybe this one just because these code is already used above) have no value specified?

src/templates/snippets/cancerStaging.js
```diff
@@ -0,0 +1,38 @@
...
+      method: {
+        coding: [
+          coding({
+            code,
+            system: 'http://snomed.info/sct',
+          }),
+        ],
+      },
+    };
+  } if (code === 'C146985') {
```
jafeltra, line 18: This is also really small, but is this meant to be an `else if` or just another `if`? I don't think it makes any real difference either way, but I think usually we have `if` statements on their own line.

src/templates/snippets/cancerStaging.js
```diff
@@ -0,0 +1,38 @@
...
+          coding({
+            code,
+            system: 'http://ncimeta.nci.nih.gov',
+          }),
+        ],
+      },
+    };
+  }
+  logger.warn(`stagingMethodTemplate received a code ${code} that is not recognized; code will not be added to the resulting FHIR resource`);
+  return null;
```
jafeltra, line 33: [JA1] It looks like the `method` value set is "extensible" so I think we might want to add codes that are provided even if they aren't one in the mCODE Value Set. I think a warning message still makes sense to log, but maybe we should return a `method` `CodeableConcept` with just the provided code.

jafeltra (commented):
This looks really great! It's so nice that all of this is contained to one repo, so yay for all the refactoring helping us out! I just had a couple comments inline and one other below:

[JA2]: Do you think we should include the new `stagingSystem` in the data we test in the staging template tests? It looks like those tests are checking the various pieces of data that can get passed into the template.

test/extractors/CSVStagingExtractor.test.js
```diff
@@ -33,6 +33,7 @@ describe('CSVStagingExtractor', () => {
       delete localData.m;
       delete localData.n;
       delete localData.type;
+      delete localData.stagingSystem;
       delete localData.stageGroup;
```
julianxcarter, line 37: JC1: This is definitely a nitpick, but maybe we should be setting these optional properties to `null` here rather than performing a `delete` operation. In the wild, the template will always receive all of the properties as an argument, so we may want to model that behavior in the tests as well.

test/sample-client-data/staging-information.csv
Dtphelan1: Yes, great idea. Will do.

src/templates/snippets/cancerStaging.js
Dtphelan1: I think our linter auto-changed this to an `if`. I don't have a strong opinion either way, just let me know if you do 😄

test/sample-client-data/staging-information.csv
Dtphelan1: Fixed in my most recent push

test/extractors/CSVStagingExtractor.test.js
Dtphelan1: That's a really good point given our new approach to empty values 👍  I'll do that now

Dtphelan1:
Most recent changes should address all @julianxcarter and @jafeltra comments - let me know what you both think of the new changes 😄

src/templates/snippets/cancerStaging.js
Dtphelan1: should be addressed now 😄

jafeltra (approved):
This looks excellent! Thanks for making those updates!

julianxcarter (approved):
This looks great! Nice work!

PR #65, opened 2020-12-08
Generic valueSet lookup

Dtphelan1:
# Summary
Implements a generic valueSet helper for loading valueSets from provided paths, replacing the numerous implementations of checkCodeInVs with a single one. Additionally, this task unearthed an issue with the mCODE condition valueSets. After raising this with Mark and May, they pushed out an updated version of valueSets with ICD-10 codes in them, now using the period separator. This PR incorporates those new valueSets.

## New behavior
Anywhere that implemented a checkCodeInVs function now uses the one defined by `valueSetUtils.js`. This file also defines an enum of possible valueSet file types supported; currently only the JSON type has a defined loader and parser. 

## Code changes
- Major changes to fixtures on the tails of the new ValueSets

# Testing guidance
- Ensure that the new tests for `valueSetUtils.js` are sufficient
- Ensure that all tests pass
- Ensure that everything runs as expected with the CLI


src/helpers/valueSetUtils.js
```diff
@@ -0,0 +1,71 @@
...
+ */
+const checkCodeInVs = (code, valueSetFilePath, typeOfVS = vsTypes.json) => {
+  const valueSet = loadVs(valueSetFilePath, typeOfVS);
+  let inVSExpansion = false;
+  let inVSCompose = false;
+  if (valueSet.expansion) {
+    // If valueSet has expansion, we only need to check these codes since everything in compose is in expansion
+    inVSExpansion = valueSet.expansion.contains.some((containsItem) => {
+      if (!code || !containsItem) return false;
+      // return code.system === containsItem.system && code === containsItem.code;
```
jafeltra, line 52: [JA2]: It looks like this commented out line might be left over from the icaredata-platform version of this function. Can it be removed?

Also, same for the one in the else block below.

src/helpers/valueSetUtils.js
```diff
@@ -0,0 +1,71 @@
...
+    });
+  } else {
+    // Checks if code is in any of the valueSet.compose.include arrays
+    inVSCompose = valueSet.compose.include.some((includeItem) => {
+      if (!code || !includeItem || !includeItem.concept) return false;
+      // return c.system === includeItem.system && includeItem.concept.map((concept) => concept.code).includes(code);
+      return includeItem.concept.map((concept) => concept.code).includes(code);
+    });
+  }
+  return inVSCompose || inVSExpansion;
```
jafeltra, line 63: I don't know if this falls into this task or not, and I can't remember if we already talked about it, but should we also check that a code does not fall into the `exclude` of a value set?

test/helpers/valueSetUtils.test.js
```diff
@@ -0,0 +1,81 @@
...
+
+  describe('loadJsonVs', () => {
+    test('should throw when not given a filepath', () => {
+      expect(() => loadJsonVs()).toThrow();
+    });
+    test('should throw when file does not exist', () => {
+      expect(() => loadJsonVs('./path/does/not/exist')).toThrow();
+    });
+    test('should load from the supplied filepath', () => {
+      const valueSetFilePath = path.resolve(__dirname, './fixtures/valueset-without-expansion.json');
```
jafeltra, line 26: JA1-continued: I think we'll also want to update the paths here if we want the tests to run correctly on Windows.

test/helpers/valueSetUtils.test.js
```diff
@@ -0,0 +1,81 @@
...
+    });
+    test('should throw an error when turtle is provided', () => {
+      expect(() => loadVs(undefined, vsTypes.turtle)).toThrow();
+    });
+    test('should throw an error when an unrecoginized type is provided', () => {
+      expect(() => loadVs(undefined, vsTypes.newType)).toThrow();
+    });
+    test('Should load a vs properly for json', () => {
+      const valueSetFilePath = path.resolve(__dirname, './fixtures/valueset-without-expansion.json');
+      expect(loadJsonVs(valueSetFilePath, vsTypes.json)).toEqual(exampleValueSet);
```
jafeltra, line 43: [JA3] I think we want to be testing `loadVs` here, not `loadJsonVs`, right?

test/helpers/valueSetUtils.test.js
```diff
@@ -0,0 +1,81 @@
...
+    });
+    test('should load from the supplied filepath', () => {
+      const valueSetFilePath = path.resolve(__dirname, './fixtures/valueset-without-expansion.json');
+      expect(loadJsonVs(valueSetFilePath)).toEqual(exampleValueSet);
+    });
+  });
+
+  describe('loadVs', () => {
+    test('should throw an error when xml type is provided', () => {
+      expect(() => loadVs(undefined, vsTypes.xml)).toThrow();
```
jafeltra, line 33: JA4: I'm not sure if this is overkill here, but do you think its worth while to check the errors we're throwing here. I think you should be able to test something like

```
expect(() => loadVs(undefined, vsTypes.xml)).toThrow('No defined valueset loader for `xml` type valuesets');
```

jafeltra (commented):
This looks good to me! I ran the client against this branch and master and all the output looks the same. I just had a few comments inline below.

src/helpers/cancerStagingUtils.js
```diff
@@ -1,11 +1,9 @@
...
-
-function checkCodeInVS(code, valueSet) {
-  return valueSet.compose.include[0].concept.map((c) => c.code).includes(code);
-}
+const path = require('path');
+const { checkCodeInVs } = require('./valueSetUtils');
 
 function isCancerStagingSystem(code) {
-  return checkCodeInVS(code, cancerStagingSystemVS);
+  const cancerStagingSystemVSPath = path.resolve(__dirname, './valueSets/ValueSet-mcode-cancer-staging-system-vs.json');
```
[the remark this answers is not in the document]
Dtphelan1: Great catch

src/helpers/valueSetUtils.js
Dtphelan1: This I'm not so sure of. It actually strikes me as a shortcoming of our current approach (maybe even a misinterpretation of what a "code" is) that we don't also supply a codeSystem in this lookup. I've left this line as a reminder of this current shortcoming, but maybe I can replace it with a proper comment?

test/helpers/valueSetUtils.test.js
Dtphelan1: Definitely. Done on most recent push

test/helpers/valueSetUtils.test.js
Dtphelan1: Done in most recent push

src/helpers/cancerStagingUtils.js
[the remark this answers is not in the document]
Dtphelan1: Should be addressed in the most recent PR

src/helpers/valueSetUtils.js
Dtphelan1: We could add this as a separate task? 👍

src/helpers/valueSetUtils.js
Dtphelan1: We should replace this with a proper comment and then make a corresponding task 👍

src/helpers/valueSetUtils.js
jafeltra: Task is added, so I'll resolve this!

test/helpers/valueSetUtils.test.js
Dtphelan1: On it!

src/helpers/valueSetUtils.js
Dtphelan1: Should be addressed

test/helpers/valueSetUtils.test.js
Dtphelan1: should be addressed

jafeltra (approved):
These changes look good to me! I think anything outstanding, including using the expanded value sets here, have corresponding Jira tasks to do later!

src/helpers/valueSetUtils.js
```diff
@@ -0,0 +1,73 @@
...
+ */
+const checkCodeInVs = (code, valueSetFilePath, typeOfVS = vsTypes.json) => {
+  const valueSet = loadVs(valueSetFilePath, typeOfVS);
+  let inVSExpansion = false;
+  let inVSCompose = false;
+  if (valueSet.expansion) {
+    // If valueSet has expansion, we only need to check these codes
+    inVSExpansion = valueSet.expansion.contains.some((containsItem) => {
+      if (!code || !containsItem) return false;
+      // NOTE: This is a technically incorrect interpretation of ValueSets;
```
kjmahalingam, line 52: DM1: Can you explain to me what this note implies? Why don't we do what it suggests, and are we sure we're okay with that? I understand that this is generally fine in our situation since the likelihood of a code matching incorrectly with an identical code in another system is unlikely, but at the same time having a system to match would be preferable. Is there a reason we don't have access to a system to match against where this is used?

kjmahalingam (commented):
This looks awesome. I just have one question/comment inline. I could be convinced that this isn't even an issue we need to address, but I just want some reasoning to explain why we don't currently support it.

kjmahalingam (approved):

src/helpers/valueSetUtils.js
Dtphelan1: yes I can explain it

PR #67, opened 2020-12-28
mCODE Radiation Procedure Extractor 

Dtphelan1:
# Summary
This PR adds a new extractor for mCODE Cancer Related Radiation Procedures.

## New behavior
The approach I took for this extractor is similar to the Epic Conditions extractor. It will get all Procedure resources using the regular FHIR Procedure extractor and then will filter that list to only include ones that have a code in the mCODE VS for Cancer Related Radiation Procedures.

## Code changes
Adds the new extractor and adds tests for the new extractor.

# Testing guidance
Since this extractor uses the FHIR services to get procedures, you'll need to test it using the epic-MEF repo. There is a PR there to help set up your config and instructions to test it out.


Dtphelan1:
I've got a tedTalk prepared and everything, but the short version is that I think the approach to context used here and in @jafeltra's PR needs removal in the short-term and reconsideration in the medium term. 

To illustrate, imagine using the MCODESurgicalProcedureExtractor and MCODERadiationProcedureExtractor in tandem, without the generic FHIRProcedureExtractor. The Context-check in the second extractor, our MCODERadiationProcedureExtractor, will check for resources of a type Procedure when deciding if the total FHIRProcedure request is necessary. Since there will be Procedures, namely SurgicalProcedures, we never perform the FHIRProcedure call. This means we won't identify any RadiationProcedures.

kjmahalingam (approved):
This looks good to me! @Dtphelan1 and I discussed that we should not currently be receiving radiation procedure results, and given that, I approve!

julianxcarter (approved):
Everything looks great to me as well, nice work! @Dtphelan1

PR #68, opened 2020-12-29
Code System matching in Value Set searches

julianxcarter:
# Summary
`checkCodeInVS` now checks that a code is found within the context of a specific code system, rather than simply checking that the code is in the value set.
## New behavior
If a value set's `include` object includes a `system` property; or if an individual code within a value set's `expansion` array contains a system, `checkCodeInVS` will now only return `true` if both the code is found **and** the provided systems match. If the code object in the value set does not have a system specified, then `checkCodeInVS` acts as it did before and returns `true` if the code is found at all.
## Code changes
1. A new `stagingCodeSystem` column has been added to the staging csv, with support for this column being added to the `CSVStagingExtractor` as well as the Staging and TNM templates.
2. `checkCodeInVS` now takes a `system` argument, only returning `true` if the code found in the provided value set also has the matching code system (unless the code found in the value set has no code system at all, in which case `true` is returned).
3. `isCancerStagingSystem` now checks based on the new `stagingCodeSystem` field in addition to the `stagingSystem` field.
4. `isConditionCodePrimary` and `isConditionCodeSecondary` now check based on system and code as well. A code system lookup table has been added to the `conditionUtils.js` file to convert systems from their `urn:oid` form to their uri form if needed.
# Testing guidance
Ensure that all tests pre-existing tests pass (particularly ones pertaining to the `Condition` and `Staging` templates) and also that the changes made to `checkCodeInVS` are logically sound. Happy testing!


kjmahalingam (commented):
Nice work! I have a few small comments in line and below but this otherwise looks great to me.

DM1: For the CVS templates Excel file, can we change the date in the name to reflect when it was last updated?

test/helpers/fixtures/valueset-with-expansion.json
```diff
@@ -24,6 +24,11 @@
       {
         "code": "C00.1",
         "display": "Malignant neoplasm of external lower lip"
+      },
+      {
```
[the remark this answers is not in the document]
julianxcarter: DM2: To check that systems _and_ codes are now being matched on by `checkCodeInVS` rather than just the code as before! In order to verify that in theists, at least one of these objects in the `expansion` array needed to have a `system` property.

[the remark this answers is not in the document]
kjmahalingam: Ah, understood!

kjmahalingam (approved):
Well done!

Dtphelan1:
Other than the conversation we had of returning false when we don't have the codeSystem instead of returning true (this erring on the side of caution is in line with direct [FHIR guidance re: Coding](https://www.hl7.org/fhir/datatypes.html#Coding) - see the section "In some cases, the system might not be known") I think this is good!

julianxcarter:
Slight change made to the behavior of `checkCodeInVS` after discussing with @Dtphelan1 for a bit yesterday. We thought that it would be best to require system matching when `checkCodeInVS` is run, rather than simply matching on the `code` if a coding in a VS is present but does not have a `system` specified. 

While `system` is an optional element on codings, the FHIR guidance which Dylan linked to above states pretty clearly that codings should include a system, and that "no useful processing" should be done on the `coding` in the event that it doesn't.

src/helpers/observationUtils.js
```diff
@@ -21,7 +21,7 @@ const vitalSignsCodeToTextLookup = {
 
 function isTumorMarker(code) {
   const tumorMarkerTestVSPath = path.resolve(__dirname, 'valueSets', 'ValueSet-mcode-tumor-marker-test-vs.json');
-  return checkCodeInVs(code, tumorMarkerTestVSPath);
+  return checkCodeInVs(code, 'http://loinc.org', tumorMarkerTestVSPath);
```
Dtphelan1, line 24: One _small_ comment: Is there a reason we chose to fix the codeSystem here, whereas in other places we fix this at the template level and pass it in as a second argument? If we could justify the uniqueness of this approach, or unify across all the uses of `checkCodeInVs`, we should be good to go 😄

julianxcarter: Tumor Marker resources _always_ have a LOINC code system as detailed [here in the Tumor Marker VS](http://hl7.org/fhir/us/mcode/ValueSet-mcode-tumor-marker-test-vs.html), because of this I thought abstracting away the system parameter in the `isTumorMarker` function made some sense since we would only ever be searching for codes with the LOINC code system if we're verifying if a resource is in fact a Tumor Marker. The use case here is different from say, searching for if a code corresponds to a Primary Cancer Condition because those codes could be apart of more than one code system [as detailed here in the Primary Cancer Condition VS](http://hl7.org/fhir/us/mcode/ValueSet-mcode-primary-or-uncertain-behavior-cancer-disorder-vs.html), necessitating a system argument since it can't be reasonably inferred.

Dtphelan1: Absolutely see that's the case for the TumorMarkerVS as-is; however [the mCODETumorMarker profile defines that valueSet as extensible](http://hl7.org/fhir/us/mcode/StructureDefinition-mcode-tumor-marker.html). Since it _could_ be modified and extended, and since we're expecting to move towards paths-to-valuesets-as-arguments in the future anyway, might we want to make this an argument as well?

src/extractors/MCODERadiationProcedureExtractor.js
```diff
@@ -10,7 +10,7 @@ function getMCODERadiationProcedures(fhirProcedures) {
   return fhirProcedures.filter((procedure) => {
     const coding = procedure.resource.code ? procedure.resource.code.coding : [];
     // NOTE: Update when checkCodeInVS checks code and system (might be able to pass in the full Coding)
```
Dtphelan1, line 12:
```suggestion
```

src/templates/snippets/cancerStaging.js
```diff
@@ -19,10 +19,10 @@ function methodTemplate({ code, system }) {
   )({ code, system });
 }
 
-function stagingMethodTemplate({ code }) {
-  if (isCancerStagingSystem(code)) {
+function stagingMethodTemplate({ code, system }) {
+  if (isCancerStagingSystem(code, system)) {
     // NOTE: our general value set lookup should probably return the associate system, since it's available at the VS level
```
Dtphelan1, line 24:
```suggestion
```

Dtphelan1:
I think this looks good to me!

Dtphelan1 (approved):
This looks good to me, but might benefit from a second look over by @kjmahalingam

kjmahalingam:
Still approved!

PR #69, opened 2021-01-05
Adverse event csv

Dtphelan1:
# Summary

Adds a template format for AdverseEvents, emphasizing use of adverse events in conjunction with clinical trials. No new behavior or tests, as this is a documentation-only PR. Review the proposed CSV format for: 
- Column names being appropriate;
- Default values proposed being appropriate;
- Required fields being appropriate;
- And the actual [AdverseEvent FHIR R4 spec](https://www.hl7.org/fhir/adverseevent.html) matching up with the proposed format. 

Note: This is currently missing an example sheet in the form of a CSV. I wanted to get feedback on the template format first so I didn't need to make changes in two places.


jafeltra:
This looks really good. I just had a few small comments:

- This is pretty small, but do you think we should include a `categoryCodeDisplayText`, `adverseEventDisplayText`, and `seriousnessDisplayText` as optional fields to include a display text on those CodeableConcepts?
- Not to continue to open this can of worms, but it looks like the `subject` field on AdverseEvents can be a reference to a few different things. I think we were assuming it was a reference to a patient and using the `mrn` value. Is that true? Or do we also need a `subjectType` field?

Dtphelan1:
1) I don't think that we expose displayText for any other references, so I'm of the opinion that we don't need to expose that at this time.
2) That's a fair point, but I think for our purposes we can start by assuming (and stipulating in the CSV template file) that the subject must refer to a Patient resource. After talking with Rob this morning, I think this is a safe approach.

julianxcarter (commented):
This looks great to me! One super small change in the Required Fields section of the excel sheet, I think we mean to have `adverseEventCode` as a required field rather than just `adverseEvent` as we currently do. Otherwise, I'm totally fine with this!

Dtphelan1:
Julian's great catch has been addressed!

docs/adverse-event.csv
```diff
@@ -0,0 +1,3 @@
+mrn,adverseEventId,adverseEvent,adverseEventCodeSystem,suspectedCauseId,suspectedCauseType,category,categoryCodeSystem,seriousness,seriousnessCodeSystem,severity,actuality,studyId,effectiveDate,recordedDate
```
jafeltra, line 1: Do you think the columns here should be updated to reflect the latest changes to the columns in the spreadsheet?

Dtphelan1: oh my gash

jafeltra (approved):
Beautiful!

julianxcarter (approved):
This looks great to me! 👍🏽

PR #73, opened 2021-01-13
Removes commented out blocks of code

Dtphelan1:
Removes a number of commented out blocks of code that leveraged the previous context strategy. Until we decide how Context ought to be used, I'm removing that commented out code.


jafeltra (approved):
Seems very reasonable!

kjmahalingam (approved):

PR #74, opened 2021-01-19
Improve error message on unsuccessful extractor initialization 

Dtphelan1:
Added try catch block for better error management in the case of extractor initialization. 

Test that the error is helpful by 1) changing the type of an extractor in your config file to be nonsense, and 2) reviewing the error message for quality. 

Also test that normal extraction works as expected. 

**_only needs one reviewer_**


julianxcarter (approved):
I really like this change! Everything works as expected here, you've got my approval 👍🏽.

PR #75, opened 2021-01-19
Update careplan to create multiple extensions

mgramigna:
# Summary

He's back baby.

Updates Care plan extractor/template to create multiple extension entries rather than multiple CarePlan resources when provided >1 entry to the extractor.

## New behavior

Multiple CarePlan data entries now appear as multiple `http://mcodeinitiative.org/codex/us/icare/StructureDefinition/icare-care-plan-review` extensions in the outputted CarePlan resource

## Code changes

* Update joinAndReformat function to produce a new type of data (each "review" maps to a new extension)
``` Javascript
        {
          mrn: 'mrn-1',
          reviews: [
            {
              effectiveDate: '2020-04-15',
              hasChanged: 'true',
              reasonCode: '281647001',
            },
            ...
          ],
        }
```
* Update tests/fixtures accordingly
* Removed the usage of the `created` property for now as it is not required and we want to avoid inferring its value based on the multiple dates that could now be present for 1 CarePlan resource

# Testing guidance

Run CLI and tests and inspect the output. The unit tests cover multiple extensions, but to see this in action, you can add a row to `test/sample-client-data/treatment-plan-change-information.csv` for one of the patients:

e.g.

``` csv
mrn,reasonCode,changed,dateOfCarePlan,dateRecorded
123,281647001,true,2020-04-15,2020-04-15
123,281647001,true,2020-04-30,2020-04-30
456,405613005,true,2020-03-30,2020-04-15
789,266721009,true,2020-04-22,2020-06-17
```

For patient 123, this will now have 1 CarePlan resource with 2 extensions rather than 2 resources in the output bundle.

### DIFF
--- a/src/extractors/CSVTreatmentPlanChangeExtractor.js
+++ b/src/extractors/CSVTreatmentPlanChangeExtractor.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const _ = require('lodash');
 const { Extractor } = require('./Extractor');
 const { CSVModule } = require('../modules');
 const { formatDate, formatDateTime } = require('../helpers/dateUtils');
@@ -8,23 +9,36 @@ const logger = require('../helpers/logger');
 
 // Formats data to be passed into template-friendly format
 function formatData(tpcData) {
-  logger.debug('Reformatting disease status data from CSV into template format');
-  return tpcData.map((data) => {
-    const { mrn, dateOfCarePlan, changed, reasonCode } = data;
-    if (!mrn || !dateOfCarePlan || !changed) {
+  logger.debug('Reformatting treatment plan change data from CSV into template format');
+
+  // If there are multiple entries, combine them into one object with multiple reviews
+  const combinedData = _.reduce(tpcData, (res, n) => {
+    if (!n.mrn || !n.dateOfCarePlan || !n.changed) {
       throw new Error('Treatment Plan Change Data missing an expected property: mrn, dateOfCarePlan, changed are required');
     }
 
     // reasonCode is required if changed flag is true
-    if (changed === 'true' && !reasonCode) {
+    if (n.changed === 'true' && !n.reasonCode) {
       throw new Error('reasonCode is required when changed flag is true');
     }
+
+    if (!res.mrn) res.mrn = n.mrn;
+    (res.reviews || (res.reviews = [])).push({
+      dateOfCarePlan: n.dateOfCarePlan,
+      reasonCode: n.reasonCode,
+      changed: n.changed,
+    });
+    return res;
+  }, {});
+
+  // Format each entry in the reviews array
+  combinedData.reviews = combinedData.reviews.map((reviews) => {
+    const { dateOfCarePlan, changed, reasonCode } = reviews;
 
     const formattedData = {
       effectiveDate: formatDate(dateOfCarePlan),
       effectiveDateTime: formatDateTime(dateOfCarePlan),
       hasChanged: changed,
-      mrn,
     };
 
     // Add reasonCode to formattedData if available
@@ -34,6 +48,9 @@ function formatData(tpcData) {
 
     return formattedData;
   });
+
+  // Array will contain one element to generate one FHIR resource with multiple extensions for reviews
+  return [combinedData];
 }
 
 class CSVTreatmentPlanChangeExtractor extends Extractor {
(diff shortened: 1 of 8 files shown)


src/extractors/CSVTreatmentPlanChangeExtractor.js
```diff
@@ -8,23 +9,36 @@ const logger = require('../helpers/logger');
 
 // Formats data to be passed into template-friendly format
 function formatData(tpcData) {
-  logger.debug('Reformatting disease status data from CSV into template format');
```
Dtphelan1, line 11: _"Where do you get off saying all we do is copy and paste code?!?"_

src/extractors/CSVTreatmentPlanChangeExtractor.js
```diff
@@ -8,23 +9,36 @@ const logger = require('../helpers/logger');
...
-  return tpcData.map((data) => {
-    const { mrn, dateOfCarePlan, changed, reasonCode } = data;
-    if (!mrn || !dateOfCarePlan || !changed) {
+  logger.debug('Reformatting treatment plan change data from CSV into template format');
+
+  // If there are multiple entries, combine them into one object with multiple reviews
+  const combinedData = _.reduce(tpcData, (res, n) => {
```
Dtphelan1, line 15: DP1: Could you talk about what `n` is shorthand for? I don't think I could gleam that from first read.

src/extractors/CSVTreatmentPlanChangeExtractor.js
```diff
@@ -8,23 +9,36 @@ const logger = require('../helpers/logger');
...
     }
 
     // reasonCode is required if changed flag is true
-    if (changed === 'true' && !reasonCode) {
+    if (n.changed === 'true' && !n.reasonCode) {
       throw new Error('reasonCode is required when changed flag is true');
     }
 
+    if (!res.mrn) res.mrn = n.mrn;
+    (res.reviews || (res.reviews = [])).push({
```
Dtphelan1, line 26: DP2: Instead of these conditionals, could we just include these properties in the seed accumulator on line 32?

Dtphelan1: Also, this doesn't need to happen on line 32. It can happen before the reduce is called. I'm just trying to make the object's structure a little more clear before looping.

src/extractors/CSVTreatmentPlanChangeExtractor.js
```diff
@@ -8,23 +9,36 @@ const logger = require('../helpers/logger');
...
+      dateOfCarePlan: n.dateOfCarePlan,
+      reasonCode: n.reasonCode,
+      changed: n.changed,
+    });
+    return res;
+  }, {});
+
+  // Format each entry in the reviews array
+  combinedData.reviews = combinedData.reviews.map((reviews) => {
+    const { dateOfCarePlan, changed, reasonCode } = reviews;
```
Dtphelan1, line 36: Could we make the the anon-fn argument `review` instead of `reviews`?

julianxcarter (approved):
This looks really good to me, especially after the slight re-org of the `_.reduce()` within the care plan extractor. I'm approving! Welcome back @mgramigna 👍🏽

src/extractors/CSVTreatmentPlanChangeExtractor.js
```diff
@@ -20,20 +34,38 @@ function formatData(tpcData) {
...
+      reasonDisplayText,
+      changed,
+    });
+    return res;
+  }, combinedFormat);
+
+  // Format each entry in the reviews array
+  combinedData.reviews = combinedData.reviews.map((review) => {
+    const { dateOfCarePlan, changed, reasonCode, reasonDisplayText } = review;
+
```
Dtphelan1, line 49: Any reason these don't happen in the same loop? It's really not the end of the world, but I wanted to see if I was missing something.

Dtphelan1:
With the exception of a possibly redundant map after the reduce, I think this looks good.

src/extractors/CSVTreatmentPlanChangeExtractor.js
mgramigna: Tbh there really isn't a good one. I kept it separate with the mindset of "oh the reformatting of the array happens first and then the formatting of the data happens after" whereas the latter looks almost exactly the same as it did before, it just acts on a different array.

We could easily combine this. LMK

Dtphelan1: That's absolutely reasonable. Let's leave it separate for now.

Dtphelan1:
LGTM

Dtphelan1 (approved):

PR #76, opened 2021-01-21
resourceType parameter added to reference snippet

julianxcarter:
# Summary
This PR adds an optional `resourceType` parameter to the reference snippet.
## New behavior
Reference-based elements on generated FHIR jsons now include the `type` property, specifying the resourceType of the artifact being referenced.
## Code changes
1. A `resourceType` parameter has been added to the reference snippet which maps directly to the `type` property on the generated reference.
2. All templates that use the reference snippet have been updated to include the `resourceType` argument. 
3. All test fixtures with a reference-based element have been updated to include the `type` property within the reference.
4. The new `resourceType` argument has been added to the `MAXIMAL_DATA` object within `reference.test.js`.
# Testing guidance
Run the MEF and review the output, ensuring that each `reference` element has a `type` property and that the value is appropriate.

### DIFF
--- a/src/templates/CancerDiseaseStatusTemplate.js
+++ b/src/templates/CancerDiseaseStatusTemplate.js
@@ -12,14 +12,14 @@ function evidenceTemplate({ evidence }) {
 function focusTemplate({ condition }) {
   return {
     focus: [
-      reference(condition),
+      reference({ ...condition, resourceType: 'Condition' }),
     ],
   };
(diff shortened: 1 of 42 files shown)


kjmahalingam (approved):
Great work!

jafeltra (approved):
Whoa great job wrangling all those changed test fixtures! This looks great!

PR #77, opened 2021-01-22
FHIR Extractor for Adverse Events

mickohanlon23:
# Summary
Adds a FHIR Extractor for Adverse Events

## New behavior
Adverse Events can now be extracted from a patient record.

#### Created FHIR Adverse Event Extractor
- A base study is hardcoded in the extractor, and added as an optional parameter. Its default value is an empty string.
- AdverseEvents have a required `subject` field instead of a `patient` field. Therefore, the patient value is copied to the subject field, and then the patient field is deleted within the scope of Adverse Event's `parametrizeArgsForFHIRModule()` function.
#### Created FHIR Adverse Event Extractor Test
- Test extractor functionality as well as extractor construction with a `study` param value
- Additional test to check for proper deletion of `patient` field

#### Added `adverse-event-infornation.csv` to `test/sample-client-data`
- Admittedly I originally thought CSV extraction was a part of my task.
- This test CSV isn't actually used to test any of the functionality in this PR, but I figure it will be useful when Adverse Events CSV extractor is implemented.

# Testing guidance

In order for the extraction to work, you must add the study ID to line 3 of the `FHIRAdverseEventExtractor.js`. Running without this will return a 400 error.

To test, add FHIRAdverseEventExtractor to the config file and edit the patient MRNs file to include a MRN of a patient with an adverse event.


test/extractors/FHIRAdverseEventExtractor.test.js
```diff
@@ -0,0 +1,64 @@
...
+
+describe('FHIRAdverseEventExtractor', () => {
+  describe('Constructor', () => {
+    test('Constructor sets resourceType as AdverseEvent', () => {
+      expect(extractor.resourceType).toEqual('AdverseEvent');
+    });
+    test('sets study based on BASE_STUDY if not provided', () => {
+      expect(extractor.study).toEqual(baseStudy);
+    });
+    test('sets status if provided', () => {
```
kjmahalingam, line 24: DM1: Should this say `'sets study if provided'` instead?

test/extractors/FHIRAdverseEventExtractor.test.js
```diff
@@ -0,0 +1,64 @@
...
+      expect(extractorWithStudy.study).toEqual(MOCK_STUDIES);
+    });
+  });
+
+  describe('parametrizeArgsForFHIRModule', () => {
+    test('should not add study when not set to param values', async () => {
+      // Create spy
```
kjmahalingam, line 31: DM2: Since we have a few tests all in this one describe, and since the spy that's being made appears to be identical per test, I think we could add this into a `beforeEach`. In case you're unfamiliar: https://jestjs.io/docs/en/setup-teardown

kjmahalingam (commented):
This looks good overall! I have a couple benign comments in-line, but otherwise this looks good to me. I haven't tested yet manually, and will do that once the changes are incorporated.

test/extractors/FHIRAdverseEventExtractor.test.js
mickohanlon23: Yes, I swear I didn't copy and paste that.

test/extractors/FHIRAdverseEventExtractor.test.js
mickohanlon23: They are actually slightly different; the 2nd and 3rd tests construct the extractor with the optional study parameter. I added a nested describe statement to combine the identical spy functionality for those two tests. Does this look like the right way to go about it?

kjmahalingam: Ah I see. Yes, this looks good to me!

kjmahalingam (approved):
Nice work!

julianxcarter (approved):
This looks great to me! Nice work 👍🏽 @mickohanlon23

PR #78, opened 2021-01-28
Added context utilities to the base-MEF

Dtphelan1:
# Summary

After realizing how many extractors will depend on getting the patient from context, this PR moves the utility from the E-MEF to this repo.  

## New behavior

Minimal, as this PR doesn't concern itself with using the functionality anywhere in this repo. 

## Code changes

- Adds this utility;
- Adds tests for the utility;
- Exports the utility.

# Testing guidance

- Make sure tests pass, especially the new ones. 
- Connect this with the E-MEF repo and ensure that tests pass there too.

PR #79, opened 2021-01-29
Updated Context Management in BaseFHIRExtractor

jafeltra:
# Summary
This PR updates the `BaseFHIRExtractor` to require the `Patient` resource to be in context when it is used to search for other resource types.

## New behavior
When any of the generic FHIR resource extractors are run before the `FHIRPatientExtractor`, an error will be thrown and that resource will not be extracted. When the `FHIRPatientExtractor` is run first, the behavior is unchanged.

## Code changes
Removed the `BaseFHIRModule` search for `Patient` when the `Patient` resource was not on the context bundle. Instead, throw an error.

Added a test in `BaseFHIRExtractor` to test for this new error. Also updated the other FHIR extractors that now needed a `Patient` on context. I did not add a test on each of the FHIR extractors to show that an error would be thrown if `Patient` was missing from context because I felt that was covered by the `BaseFHIRExtractor` test, but I can add it if reviewers feel it's helpful.

# Testing guidance
Test that the changes to the tests make sense and that the new tests cover the new context requirement. You can also manually run extraction using the E-MEF to test that everything runs the same if `FHIRPatientExtractor` is used first and that errors are thrown if it is used after other FHIR extractors in your config.

**Note**: This PR will need to be udpated if #78 is merged first.

### DIFF
--- a/src/extractors/BaseFHIRExtractor.js
+++ b/src/extractors/BaseFHIRExtractor.js
@@ -20,19 +20,18 @@ class BaseFHIRExtractor extends Extractor {
     this.baseFHIRModule.updateRequestHeaders(newHeaders);
   }
 
-  // Use mrn to get PatientId by default; common need across almost all extractors
-  async parametrizeArgsForFHIRModule({ mrn, context }) {
+  /* eslint-disable class-methods-use-this */
+  // Use context to get PatientId by default; common need across almost all extractors
+  async parametrizeArgsForFHIRModule({ context }) {
     const idFromContext = parseContextForPatientId(context);
-    if (idFromContext) return { patient: idFromContext };
-
-    logger.debug('No patient ID in context; fetching with baseFHIRModule');
-    const patientResponseBundle = await this.baseFHIRModule.search('Patient', { identifier: `MRN|${mrn}` });
-    if (!patientResponseBundle || !patientResponseBundle.entry || !patientResponseBundle.entry[0] || !patientResponseBundle.entry[0].resource) {
-      logger.error(`Could not get a patient ID to cross-reference for ${this.resourceType}`);
-      return {};
+    if (idFromContext) {
+      logger.debug('Patient found in context');
+      return { patient: idFromContext };
     }
-    return { patient: patientResponseBundle.entry[0].resource.id };
+
+    throw new Error('BaseFHIRExtractor could not find Patient resource in context. Use an extractor to get a Patient resource first.');
   }
+  /* eslint-enable class-methods-use-this */
 
   // Since different superclasses of the baseFHIRExtractor will parse the `get`
   // arguments differently, all pass to this function which interfaces with the baseFHIRModule
--- a/src/extractors/BaseFHIRExtractor.js
+++ b/src/extractors/BaseFHIRExtractor.js
@@ -1,12 +1,8 @@
 const { Extractor } = require('./Extractor');
 const { BaseFHIRModule } = require('../modules');
-const { determineVersion, mapFHIRVersions, isBundleEmpty, getBundleResourcesByType } = require('../helpers/fhirUtils');
+const { determineVersion, mapFHIRVersions, isBundleEmpty } = require('../helpers/fhirUtils');
+const { getPatientFromContext } = require('../helpers/contextUtils');
 const logger = require('../helpers/logger');
-
-function parseContextForPatientId(context) {
-  const patientInContext = getBundleResourcesByType(context, 'Patient', {}, true);
-  return patientInContext ? patientInContext.id : undefined;
-}
 
 class BaseFHIRExtractor extends Extractor {
   constructor({ baseFhirUrl, requestHeaders, version, resourceType }) {
@@ -22,14 +18,9 @@ class BaseFHIRExtractor extends Extractor {
 
   /* eslint-disable class-methods-use-this */
   // Use context to get PatientId by default; common need across almost all extractors
-  async parametrizeArgsForFHIRModule({ context }) {
-    const idFromContext = parseContextForPatientId(context);
-    if (idFromContext) {
-      logger.debug('Patient found in context');
-      return { patient: idFromContext };
-    }
-
-    throw new Error('BaseFHIRExtractor could not find Patient resource in context. Use an extractor to get a Patient resource first.');
+  async parametrizeArgsForFHIRModule({ mrn, context }) {
+    const patient = getPatientFromContext(mrn, context);
+    return { patient: patient.id };
   }
   /* eslint-enable class-methods-use-this */
 
--- a/src/extractors/BaseFHIRExtractor.js
+++ b/src/extractors/BaseFHIRExtractor.js
@@ -18,6 +18,7 @@ class BaseFHIRExtractor extends Extractor {
 
   /* eslint-disable class-methods-use-this */
   // Use context to get PatientId by default; common need across almost all extractors
+  // NOTE: Async because other extractors that extend this may need to make async lookups in the future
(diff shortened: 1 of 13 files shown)


src/extractors/BaseFHIRExtractor.js
```diff
@@ -20,19 +20,18 @@ class BaseFHIRExtractor extends Extractor {
...
-    if (idFromContext) return { patient: idFromContext };
-
-    logger.debug('No patient ID in context; fetching with baseFHIRModule');
-    const patientResponseBundle = await this.baseFHIRModule.search('Patient', { identifier: `MRN|${mrn}` });
-    if (!patientResponseBundle || !patientResponseBundle.entry || !patientResponseBundle.entry[0] || !patientResponseBundle.entry[0].resource) {
-      logger.error(`Could not get a patient ID to cross-reference for ${this.resourceType}`);
-      return {};
+    if (idFromContext) {
+      logger.debug('Patient found in context');
+      return { patient: idFromContext };
```
Dtphelan1, line 29: DP1 - One thing I did on my PR which I might recommend here is reversing the if-block logic here; I think we want to conditionally throw an error if there does not exist an `idFromContext`, and assume we can safely return the `{ patient: idFromContext }` field if we have an `idFromContext`. Let me know if that makes sense!

test/extractors/BaseFHIRExtractor.test.js
```diff
@@ -66,16 +66,14 @@ describe('BaseFhirExtractor', () => {
...
   });
 
-  test('parametrizeArgsForFHIRModule makes Patient call if context has no relevant ID', async () => {
+  test('parametrizeArgsForFHIRModule throws an error if context has no relevant ID', async () => {
     baseFHIRModuleSearchSpy.mockClear();
-    const paramsBasedOnContext = await baseFHIRExtractor.parametrizeArgsForFHIRModule({ mrn: MOCK_PATIENT_MRN });
-    expect(baseFHIRModuleSearchSpy).toHaveBeenCalled();
-    expect(paramsBasedOnContext).toHaveProperty('patient');
-    expect(paramsBasedOnContext.patient).toEqual(examplePatientBundle.entry[0].resource.id);
+    await expect(baseFHIRExtractor.parametrizeArgsForFHIRModule({ mrn: MOCK_PATIENT_MRN })).rejects.toThrowError('BaseFHIRExtractor could not find Patient resource in context.');
```
Dtphelan1, line 71: NOTE: this will have to change when you start using my helper function. I leave it to you to decide if we want an explciit error message or not based no our conversations earlier. 😄

src/extractors/BaseFHIRExtractor.js
```diff
@@ -20,19 +20,18 @@ class BaseFHIRExtractor extends Extractor {
     this.baseFHIRModule.updateRequestHeaders(newHeaders);
   }
 
-  // Use mrn to get PatientId by default; common need across almost all extractors
-  async parametrizeArgsForFHIRModule({ mrn, context }) {
+  /* eslint-disable class-methods-use-this */
+  // Use context to get PatientId by default; common need across almost all extractors
+  async parametrizeArgsForFHIRModule({ context }) {
```
Dtphelan1, line 25: DP2 - It looks like we've removed the mrn from these function calls. Do we want to remove it from the function mocks below?

test/extractors/MCODERadiationProcedureExtractor.test.js
```diff
@@ -18,44 +18,6 @@ const fhirProcedureExtractorSpy = jest.spyOn(fhirProcedureExtractor, 'get');
 
 describe('MCODERadiationProcedureExtractor', () => {
   describe('getFHIRProcedures', () => {
-    // it('should return procedure entries for patient from context', async () => {
```
Dtphelan1, line 21: Question: are we punting on this until we take on these tasks officially?

src/extractors/BaseFHIRExtractor.js
jafeltra: That makes sense! I think I might just get rid of the `parseContextForPatientId` function too or rework it once I use your function.

src/extractors/BaseFHIRExtractor.js
jafeltra: Ohh good call. That was definitely an oversight that I'll fix!

test/extractors/BaseFHIRExtractor.test.js
jafeltra: An exercise for the reader!

test/extractors/MCODERadiationProcedureExtractor.test.js
jafeltra: Umm good question. I think I took them out because I thought the extractor was working according to how we want it to now and follows the right context pattern. But if it doesn't, I can either add in a test, put back the comments, or make a new task because I don't think we have tasks for these two extractors.

jafeltra:
DP2 should be resolved. I think the other two comments will need to wait for your PR to get merged. Unless I branch off of it or add it myself.

test/extractors/BaseFHIRExtractor.test.js
Dtphelan1: hehehe

src/extractors/BaseFHIRExtractor.js
Dtphelan1: Nice! Let's also make sure that we make that change in other relevant extractors test's

Dtphelan1: Alternatively, if we add back the MRN as the getPatientFromContext function expects, then we won't have to make those changes. Based on our recent conversation, I think this is the approach we will take.

jafeltra:
DP1 - I removed the logic all together and just depend on the `getPatientFromContext` function to throw the error, similarly to the Cancer Disease Status extractor.
DP2 - I added back the MRN and reverted the tests to continue including it. 
I have also done the exercise as the reader and just tested that an error was thrown.

Let me know if you notice anything else!

src/extractors/BaseFHIRExtractor.js
```diff
@@ -20,19 +16,13 @@ class BaseFHIRExtractor extends Extractor {
     this.baseFHIRModule.updateRequestHeaders(newHeaders);
   }
 
-  // Use mrn to get PatientId by default; common need across almost all extractors
+  /* eslint-disable class-methods-use-this */
+  // Use context to get PatientId by default; common need across almost all extractors
```
Dtphelan1, line 20: DP3 - Do we want to make a comment here similar to what we do in the baseclient saying that we make this async anticipatorily?

test/extractors/FHIRAdverseEventExtractor.test.js
```diff
@@ -1,14 +1,22 @@
...
+  resourceType: 'Bundle',
+  entry: [
+    {
+      fullUrl: 'context-url',
+      resource: { resourceType: 'Patient', id: MOCK_MRN },
+    },
+  ],
+};
+
+// Construct extractor and create spies for mocking responses
```
Dtphelan1, line 19: nice typo fix

jafeltra: I was already making changes in the file, saw it, and had to do it.

src/extractors/BaseFHIRExtractor.js
jafeltra: Should be added in 94df556!

Dtphelan1 (approved):

julianxcarter (approved):
Looks good to me!

PR #80, opened 2021-02-04
CSV Adverse Event Extractor

julianxcarter:
# Summary
This PR adds a new `CSVAdverseEventExtractor`, and also adds handy `ifSomeArgsArr` and `ifAllArgsArr` helper functions within `templateUtils` that may be useful to us in the future. Big shoutout to @mickohanlon23 who did the majority of this work!
## New behavior
The MEF now extracts `AdverseEvent` resources from CSVs.
## Code changes
1. The `CSVAdverseEventExtractor` and `AdverseEventTemplate` have been created.
2. Test file and corresponding fixtures have been created for the `CSVAdverseEventExtractor` and `AdverseEventTemplate`.
3. Two new functions have been added to `templateUtils.js`,  `ifSomeArgsArr` and `ifAllArgsArr`. `ifSomeArgsArr` was created pretty much exclusively to go hand in hand with the approach that I took to formatting the `category` data that we pull from the CSV. When I created that, I thought that it'd make sense to add an `ifAllArgsArr` function as well.
4. Unit tests for `ifSomeArgsArr` and `ifAllArgsArr` have been created as well.
# Testing guidance
Add the following to your config file and then run the client from the MEF, ensuring that the output looks correct, test cases pass, and also that the tests look sufficient: 
```
    {
      "label": "adverseEvent",
      "type": "CSVAdverseEventExtractor",
      "constructorArgs": {
        "filePath": "./test/sample-client-data/adverse-event-information.csv"
      }
    }
```
Happy testing!


src/extractors/CSVAdverseEventExtractor.js
```diff
@@ -0,0 +1,73 @@
...
+      category, categoryCodeSystem, categoryDisplayText, severity, actuality, studyId, effectiveDate, recordedDate,
+    } = data;
+
+    if (!(mrn && adverseEventCode && effectiveDate)) {
+      throw new Error('The adverse event is missing an expected attribute. Adverse event code, mrn, and effective date are all required.');
+    }
+
+    const categoryCodes = category.split('|');
+    const categorySystems = categoryCodeSystem.split('|');
+    const categoryDisplays = categoryDisplayText.split('|');
```
Dtphelan1, line 23: DP1: Do we want to enforce a cardinality-match with these values? I can imagine there will be difficulties inferring how to distribute the values otherwise.

Dtphelan1:
All of the helper functions for the templates look good to me. The tests look solid there 👍

src/extractors/CSVAdverseEventExtractor.js
julianxcarter: DP1: I think that'd be a great idea, it does introduce some decisions to make regarding how we handle the `codeSystem` and `displayText` values given that they're optional. I think a sensible solution would be that if more than one `categoryCode` is included, then we necessitate each code to have a corresponding `categoryCodeSystem` and enforce cardinality matching there but still allow for there to be fewer `displayText` values (matching them sequentially in that case). 

Another solution could be to enforce cardinality matching for all three value, but necessitate empty strings (e.g. `''`) to be included between each pipe so we know if a value was intentionally excluded. That would ensure that we're always matching the three values correctly, but would also ask a bit more of our users. Open to talking about this more to see which seems like the most palatable option!

Dtphelan1: I think I lean towards the last option personally - it mirrors how empty values need to be defined in CSV files generally. I'd propose we move forward with that, unless there are any concerns about that approach 😄

julianxcarter: That sounds good to me! Adding it in now

Dtphelan1:
DP2: It might be nice to include the AdverseEvent configuration in the `mcode-csv-config.example.json` starter file.

Dtphelan1:
DP3: I think the sample-data doesn't currently work with the cardinality matching checks...

Dtphelan1:
Also DP4: Could we add an example of the multiple-category AdverseEvents to the `adverse-events.csv` file?

julianxcarter:
DP3 and DP4 have both been addressed! @Dtphelan1

Dtphelan1:
I think the structure of the AdverseEvent resource looks good - reviewed it manually and compared it to the example resource on FHIR.org.

Dtphelan1 (approved):
After reviewing the AdverseEvent template file and the tests, I'm good to approve! Great work @julianxcarter  (and @mickohanlon23)!

src/extractors/CSVAdverseEventExtractor.js
```diff
@@ -0,0 +1,77 @@
...
+
+    if (!(categoryCodes.length === categorySystems.length && categoryCodes.length === categoryDisplays.length)) {
+      throw new Error('A category atrribute on the adverse event is missing a corresponding categoryCodeSystem or categoryDisplayText value.');
+    }
+
+
+    return {
+      id: adverseEventId,
```
jafeltra, line 31: I'm not entirely positive about this one, but it doesn't look like `adverseEventId` is specified as required, so I think we might want to generate it if it doesn't exist. It looks like at least the `CSVCancerRelatedMedicationExtractor` supports something like this by doing

```
...(medicationId && { id: medicationId }),
```
Do you think we should do that here? Was that the intention when you made the spreadsheet @Dtphelan1? It looks like maybe other CSV files have made this id required, so we could also go that route and then the code here won't change.

jafeltra (commented):
This looks really great! I just had a few, pretty minor comments below, and one question that I'm happy to chat more about because I don't know the right answer.

docs/adverse-event.csv
```diff
@@ -1,3 +1,4 @@
 mrn,adverseEventId,adverseEventCode,adverseEventCodeSystem,adverseEventDisplayText,suspectedCauseId,suspectedCauseType,seriousness,seriousnessCodeSystem,seriousnessDisplayText,category,categoryCodeSystem,categoryDisplayText,severity,actuality,studyId,effectiveDate,recordedDate
 mrn-full-example,example-id-1,event-code,code-system,code-display,cause-id,resourceType,seriousness-code,code-system,seriousness-display,category-code,code-system,category-dislpay,mild,actual,id,12-09-1994,12-09-1994
+mrn-two-category-example,example-id-1,event-code,code-system,code-display,cause-id,resourceType,seriousness-code,code-system,seriousness-display,category-code|category-code,code-system|code-system,category-display|category-display,mild,actual,id,12-09-1994,12-09-1994
 mrn-minimal-example,,code-from-default-system,,,,,,,,,,,,,,12-09-1994,
```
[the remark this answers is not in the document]
julianxcarter: Absolutely! I'll swap these dates to keep things consistent, good catch!

src/extractors/CSVAdverseEventExtractor.js
```diff
@@ -0,0 +1,77 @@
...
+      subjectId: mrn,
+      code: adverseEventCode,
+      system: !adverseEventCodeSystem ? 'http://snomed.info/sct' : adverseEventCodeSystem,
+      display: adverseEventDisplayText,
+      suspectedCauseId,
+      suspectedCauseType,
+      seriousnessCode: seriousness,
+      seriousnessCodeSystem,
```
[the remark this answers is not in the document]
julianxcarter: I accounted for the default value [in the template rather than the extractor](https://github.com/mcode/mcode-extraction-framework/blob/8fa8ec10dd16b086eb4ba41921a6ca0103475cf3/src/templates/AdverseEventTemplate.js#L30). I'm actually not at all sure why I chose to do it that way though so I'll do it in the extractor instead so that it aligns with our approach in other extractors!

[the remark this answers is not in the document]
jafeltra: Oh wait, I totally missed that! I'm good with either way!

docs/adverse-event.csv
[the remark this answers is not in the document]
Dtphelan1: Dang this is surely my fault...

src/extractors/CSVAdverseEventExtractor.js
julianxcarter: I definitely think that we should do what you suggested here since the `adverseEventId` field isn't specified as required. 

I'm not sure if there's something specific to the use case of AdverseEvent or MedicationStatement resources that make us not require `id` fields on those CSVs, but it does appear that the `id` field has a cardinality of 0 to 1 on just about all of the FHIR profiles that we support. Maybe this should be a blanket change?

Dtphelan1: That looks good to me!

jafeltra: I think at least for the Condition CSV, we require the `conditionId` because we reference it in other CSVs like CancerDiseaseStatus. But it doesn't look like we consistently make that distinction. I don't think you need to worry about making everything consistent now, and we can just update the `adverseEventId` one for now!

Dtphelan1: ^

src/extractors/CSVAdverseEventExtractor.js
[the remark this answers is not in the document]
Dtphelan1: I think we should definitely move it into the extractor for now instead of the template, if that's not a lot of work 😄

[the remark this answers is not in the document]
julianxcarter: Consistency is good! I moved the default value to the extractor in the most recent commit 🙂

jafeltra (approved):
This looks so great! Thanks for making those changes!

PR #81, opened 2021-02-10
Conditions from Context Util

jafeltra:
# Summary
This PR adds a util function for getting Condition entries from context

## New behavior
Adds a new util function, but no new behavior since it is not yet used in this repo.

## Code changes
Add the function and tests for the function

# Testing guidance
Check that the tests are sufficient for the function. The use of this function will really be tested in the related PR in the Epic-MEF (there are instructions in that PR).


mgramigna (approved):
Looks good to me. Tested everything and the approach makes sense

kjmahalingam (approved):
💯

PR #82, opened 2021-02-10
Creates a getConditionsFromContext function for parsing Condition resources off Context

Dtphelan1:
# Summary
This PR adds a util function for getting Condition resources from context.

## New behavior
Adds a new util function, but no new behavior since it is not yet used in this repo.

## Code changes
- Removes the MRN parameter from both ConditionContextUtils; it's late and I'd be happy to talk about reverting this tomorrow, but it felt right to avoid this habit of stubbing out variables we might get around to using in the future. Rob was right, as usual. 
- Add the function and tests for the function
- Added JS Docs for both condition functions in `contextUtils.js`
- Minor variable name changes in getConditionEntriesFromContext.

# Testing guidance
Check that the tests are sufficient for the function. The use of this function will really be tested in the related PR in the Epic-MEF (there are instructions in that PR).

### DIFF
--- a/test/helpers/contextUtils.test.js
+++ b/test/helpers/contextUtils.test.js
@@ -1,4 +1,4 @@
-const { getConditionEntriesFromContext, getEncountersFromContext, getPatientFromContext } = require('../../src/helpers/contextUtils');
+const { getConditionEntriesFromContext, getConditionsFromContext, getEncountersFromContext, getPatientFromContext } = require('../../src/helpers/contextUtils');
 
 const MOCK_PATIENT_MRN = '123';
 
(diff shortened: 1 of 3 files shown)


jafeltra (commented):
Oh no. I'm sorry I didn't go back and name my variables more carefully once I realized I needed the entries :(

Is there a reason for removing `mrn` from the two condition functions and not the patient function?

mgramigna:
this is egregious

Dtphelan1:
I think it's fair to remove MRN from the patient filter as well. I think for now we just need to assume that a given context-bundle corresponds to data for a given patient.

jafeltra:
> I think it's fair to remove MRN from the patient filter as well. I think for now we just need to assume that a given context-bundle corresponds to data for a given patient.

Alrighty! It might affect a fair number of files if we take it out of the arguments for `parametrizeArgsForFHIRModule`, but I think that can be a Monday decision, or at least a Monday implementation 😁

Dtphelan1:
Just flagging that I will be adding those changes this afternoon.

jafeltra:
I'm honestly not entirely convinced it's worth it, so we can talk about it first!

Dtphelan1:
I have made the changes necessary to remove the mrn from our calls to `parameterizeArgs` – while it was minimal effort, it clutters the PR aggressively. I've separated it into a separate PR; once we approve this one, we can rebase that other PR onto the new master and review it. I think that should only take one person, so that shouldn't impact much.

test/helpers/contextUtils.test.js
```diff
@@ -27,7 +27,42 @@ describe('getPatientFromContext', () => {
...
+      },
+    ],
+  };
+  test('Should return Condition resource in context', () => {
+    const conditions = getConditionsFromContext(conditionContext);
+    expect(conditions).toContain(conditionResource);
+    expect(conditions).toContain(conditionResource2);
+  });
+
+  test('Should throw an error if there is no patient in context', () => {
```
jafeltra, line 59: Really small suggestion
```suggestion
  test('Should throw an error if there is no Condition resource in context', () => {
```

jafeltra (commented):
This looks good! Good idea to separate out the `mrn` changes into a different PR. I just had one really minor comment in a test description.

jafeltra (approved):
Looks great!

mgramigna (approved):

PR #84, opened 2021-02-16
Add identifier helper and update ResearchSubject template

mgramigna:
# Summary

The motivation for this was to include the system of the ResearchStudy identifier that now gets searched for in the Epic extractor. We were always defaulting this to an example.com system. Now that we will sometimes have the system from the actual resource we search for, we want to pass this through.

## New behavior

If using Epic extractor and find a ResearchStudy resource via an API call, the system will persist over to the `study` attribute of the corresponding  `ResearchSubject` resource that is generated using the template

## Code changes

* ResearchSubject template now includes a defaulted arg for the identifier system of the ResearchStudy.
* Add fn in fhirUtils that grabs first identifier (null if not present), and corresponding unit tests

# Testing guidance

Ensure current behavior in base MEF is unchanged. More thorough testing can be done on corresponding Epic-MEF PR.

### DIFF
--- a/src/helpers/fhirUtils.js
+++ b/src/helpers/fhirUtils.js
@@ -95,6 +95,13 @@ function mapFHIRVersions(resource, currentVersion, targetVersion) {
   return resource;
 }
 
+function firstIdentifierEntry(resource) {
+  if (resource.identifier && resource.identifier.length > 0) {
+    return resource.identifier[0];
+  }
+  return null;
+}
+
 // Utility function to get the resources of a type from our bundle
 // Optionally get only the first resource of that type via 'first' parameter
 const getBundleResourcesByType = (bundle, type, context = {}, first = false) => {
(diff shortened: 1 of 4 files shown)


julianxcarter (approved):
LGTM!

test/helpers/fhirUtils.test.js
```diff
@@ -43,8 +44,8 @@ describe('isBundleEmpty', () => {
 
 describe('firstEntryInBundle', () => {
   test('Empty bundle should return undefined', () => {
+    expect(firstEntryInBundle(emptyBundle)).toBeUndefined();
```
Dtphelan1, line 47: 👀

src/helpers/fhirUtils.js
```diff
@@ -95,6 +95,13 @@ function mapFHIRVersions(resource, currentVersion, targetVersion) {
   return resource;
 }
 
+function firstIdentifierEntry(resource) {
+  if (resource.identifier && resource.identifier.length > 0) {
+    return resource.identifier[0];
+  }
+  return null;
+}
+
```
Dtphelan1, line 104: This makes me happy

Dtphelan1:
Not more work for now, but have we created a task for updating the CSV format to include this new System field? We should create a JIRA to do that, and priotize it sometime later this week.

Dtphelan1 (approved):
This looks good to me

jafeltra:
> Not more work for now, but have we created a task for updating the CSV format to include this new System field? We should create a JIRA to do that, and prioritize it sometime later this week.

I don't believe there is a Jira task for this, but I can create one!

PR #85, opened 2021-02-17
Removing the use of MRN in the parametrize function

Dtphelan1:
**NOTE: I've made the base-branch the current staging-context PR for the sake of readability; this needs to change before merging**

PR should remove any references to the MRN in calls to the `parametrizeArgsForFHIRModule` function. 

Testing should CTRL+F around for any missed instances of vestigial MRN's, and testing should ensure that (you guessed it) tests pass.

### DIFF
--- a/src/extractors/BaseFHIRExtractor.js
+++ b/src/extractors/BaseFHIRExtractor.js
@@ -19,8 +19,8 @@ class BaseFHIRExtractor extends Extractor {
   /* eslint-disable class-methods-use-this */
   // Use context to get PatientId by default; common need across almost all extractors
   // NOTE: Async because other extractors that extend this may need to make async lookups in the future
-  async parametrizeArgsForFHIRModule({ mrn, context }) {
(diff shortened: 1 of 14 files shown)


jafeltra (commented):
This looks good! Thanks for cleaning this up! I had two comments and GitHub didn't let me comment on the line I wanted to:

JA1 - do we still need the `mrn` in the `baseFHIRExtractor.get` call on [line 59 in BaseFHIRExtractor.test.js](https://github.com/mcode/mcode-extraction-framework/pull/85/files#diff-bbbd8c528c8851fe4835d7d2bc4828ceb9917e2484d4e0f0c7c30e618eb2c24aR59)?
JA2 - I think we'll need to update the few places we use `getPatientFromContext` in Epic-MEF so that the function calls work next time we update the MEF version there.

Let me know what you think!

jafeltra (approved):
Makes sense! JA1 is fixed and there's a PR for JA2!

PR #86, opened 2021-02-24
Parsing of 'not evaluated' status value in CDS extractor

julianxcarter:
# Summary
The CancerDiseaseStatusExtractor now excludes a `valueCodeableConcept` value from the resulting resource when a value corresponding to `not evaluated` is included in the diseaseStatusCode field.
## New behavior
When `709137006` (the code corresponding to `not evaluated`) is included in the diseaseStatusCode column of a CDS csv file, the CancerDiseaseStatusExtractor now populates the valueCodeableConcept element with a `dataAbsentReason` extension with a value of `not-asked`.
## Code changes
1. If the `value` object that is passed into the CancerDiseaseStatusTemplate from the extractor includes a `code` value of `709137006`, the template generates the `valueCodeableConcept` property with a `dataAbsentReason` extension.
2. The minimal extraction test case within the CancerDiseaseStatus template test file has been updated to test this new extraction scenario, with the `value` object corresponding to `not evaluated` and the resulting bundle having the `dataAbsentReason` extension.
# Testing guidance
Ensure that the updates to the CancerDiseaseStatusTemplate and its test file make sense. Also test that when including a `diseaseStatusCode` of '709137006' in the sample csv file, the resulting resource looks correct to you. Happy testing!

### DIFF
--- a/src/extractors/CSVCancerDiseaseStatusExtractor.js
+++ b/src/extractors/CSVCancerDiseaseStatusExtractor.js
@@ -16,15 +16,14 @@ class CSVCancerDiseaseStatusExtractor {
     logger.debug('Reformatting disease status data from CSV into template format');
     // Check the shape of the data
     arrOfDiseaseStatusData.forEach((record) => {
-      if (!record.mrn || !record.conditionId || (!record.diseaseStatusCode && record.observationStatus !== 'not evaluated') || !record.dateOfObservation) {
+      if (!record.mrn || !record.conditionId || !record.diseaseStatusCode || !record.dateOfObservation) {
         throw new Error('DiseaseStatusData missing an expected property: mrn, conditionId, diseaseStatusCode, and dateOfObservation are required.');
       }
     });
     const evidenceDelimiter = '|';
     return arrOfDiseaseStatusData.map((record) => ({
       status: record.observationStatus || 'final',
-      // If the Disease Status was not evaluated, there will be no value to make a record of and this property will be null
-      value: record.observationStatus === 'not evaluated' ? null : {
+      value: {
         code: record.diseaseStatusCode,
         system: 'http://snomed.info/sct',
         display: record.diseaseStatusText ? record.diseaseStatusText : getDiseaseStatusDisplay(record.diseaseStatusCode, this.implementation),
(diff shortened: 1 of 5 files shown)


src/templates/CancerDiseaseStatusTemplate.js
```diff
@@ -1,4 +1,10 @@
-const { coding, extensionArr, reference, valueX } = require('./snippets');
+const {
+  coding,
+  extensionArr,
```
Dtphelan1, line 3:
```suggestion
  extensionArr, // inline change looks like this
```

kjmahalingam (approved):
Well done! Looks great to me and works as expected.

src/extractors/CSVCancerDiseaseStatusExtractor.js
```diff
@@ -16,7 +16,7 @@ class CSVCancerDiseaseStatusExtractor {
     logger.debug('Reformatting disease status data from CSV into template format');
     // Check the shape of the data
     arrOfDiseaseStatusData.forEach((record) => {
-      if (!(record.mrn && record.conditionId && record.diseaseStatusCode && record.dateOfObservation)) {
+      if (!record.mrn || !record.conditionId || !record.diseaseStatusCode || !record.dateOfObservation) {
```
dmendelowitz, line 19: I think this change is functionally equivalent to what it was before, so it totally works, just curious if there was nay particular reason for the change?

dmendelowitz (approved):
Looks good, everything worked as expected

src/extractors/CSVCancerDiseaseStatusExtractor.js
julianxcarter: There's no particular reason. In previous commits I thought the task was something else entirely, so I had to switch up the if-statement. When I realized what the task actually was, I deleted what I had added to the if-statement but just kind of kept this reworking of it. I can switch it back if we think the old way was cleaner!

dmendelowitz: No reason to go back and change it, just wanted to make sure I wasn't missing something!

PR #87, opened 2021-02-26
Added optional trialResearchSystem column to ClinicalTrialInformationExtractor

dmendelowitz:
# Summary
ClinicalTrialInformationExtractor can now take an optional column for trialResearchSystem
## New behavior
ClinicalTrialInformationExtractor can now take an extra column for trialResearchSystem in it's CSV input. If the column is omitted from the CSV, the "system" field in the resulting study resource will also be omitted. The value can also be blank in any individual row of the CSV and the "system" field in the study resource will also be omitted
## Code changes
1. Changed CSVClinicalTrialInformationExtractor.js to take the new field, defaulting to no value if it isn't provided. Since the field is optional, it won't throw an error if it is missing
2. Changed ResearchStudyTemplate.js and ResearchSubjectTemplate.js to take in trialReseachSystem as a parameter, previously they just had hard coded default values
3. Changed CSVClinicalTrialInformationExtractor.test.js and associated files to reflect changes described above
# Testing guidance
I've changed the sample data csv (clinical-trial-information.csv) to have values for trialResearchSystem for each sample patient, but as mentioned above, any row can have a blank value for this or the entire column can be omitted.

### DIFF
--- a/src/extractors/CSVClinicalTrialInformationExtractor.js
+++ b/src/extractors/CSVClinicalTrialInformationExtractor.js
@@ -26,7 +26,9 @@ class CSVClinicalTrialInformationExtractor extends Extractor {
 
   joinClinicalTrialData(patientId, clinicalTrialData) {
     logger.debug('Reformatting clinical trial data from CSV into template format');
-    const { trialSubjectID, enrollmentStatus, trialResearchID, trialStatus } = clinicalTrialData;
+    const {
+      trialSubjectID, enrollmentStatus, trialResearchID, trialStatus, trialResearchSystem = '',
+    } = clinicalTrialData;
     const { clinicalSiteID } = this;
 
     if (!(patientId && clinicalSiteID && trialSubjectID && enrollmentStatus && trialResearchID && trialStatus)) {
@@ -40,11 +42,13 @@ class CSVClinicalTrialInformationExtractor extends Extractor {
         trialSubjectID,
         trialResearchID,
         patientId,
+        trialResearchSystem,
       },
       formattedDataStudy: {
         trialStatus,
         trialResearchID,
         clinicalSiteID,
+        trialResearchSystem,
       },
     };
   }
--- a/src/extractors/CSVClinicalTrialInformationExtractor.js
+++ b/src/extractors/CSVClinicalTrialInformationExtractor.js
@@ -26,17 +26,14 @@ class CSVClinicalTrialInformationExtractor extends Extractor {
 
   joinClinicalTrialData(patientId, clinicalTrialData) {
     logger.debug('Reformatting clinical trial data from CSV into template format');
-    let {
-      trialSubjectID, enrollmentStatus, trialResearchID, trialStatus, trialResearchSystem,
-    } = clinicalTrialData;
+    const { trialSubjectID, enrollmentStatus, trialResearchID, trialStatus } = clinicalTrialData;
+    // since trialResearchSystem is optional, check for blank value and replace with default value
+    const trialResearchSystem = (clinicalTrialData.trialResearchSystem === '') ? 'http://example.com/clinicaltrialids' : clinicalTrialData.trialResearchSystem;
     const { clinicalSiteID } = this;
 
     if (!(patientId && clinicalSiteID && trialSubjectID && enrollmentStatus && trialResearchID && trialStatus)) {
       throw new Error('Clinical trial missing an expected property: patientId, clinicalSiteID, trialSubjectID, enrollmentStatus, trialResearchID, and trialStatus are required.');
     }
-
-    //since trialResearchSystem is optional, check for blank value and replace with default value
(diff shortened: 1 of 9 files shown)


src/extractors/CSVClinicalTrialInformationExtractor.js
```diff
@@ -26,7 +26,9 @@ class CSVClinicalTrialInformationExtractor extends Extractor {
 
   joinClinicalTrialData(patientId, clinicalTrialData) {
     logger.debug('Reformatting clinical trial data from CSV into template format');
-    const { trialSubjectID, enrollmentStatus, trialResearchID, trialStatus } = clinicalTrialData;
+    const {
+      trialSubjectID, enrollmentStatus, trialResearchID, trialStatus, trialResearchSystem = '',
```
julianxcarter, line 30: JC1: Setting a default value for `trialResearchSystem` was a great call, but it's actually not needed here! When the CSVModule encounters a field with no value, it actually returns that value as an empty string rather than as an undefined variable as you may think. Because of this, we tend to not set default values for optional columns when we destructure these variables. 

A good example of this would be [in the CSVConditonExtractor](https://github.com/mcode/mcode-extraction-framework/blob/9dedb04b7adc9f6c6fb75c44f7354ab8d01597b8/src/extractors/CSVConditionExtractor.js#L13). There, you'll notice that only 5 of the fields passed in from the CSVModule are required, but the optional fields are still pulled in without default values.

src/extractors/CSVClinicalTrialInformationExtractor.js
```diff
@@ -40,11 +42,13 @@ class CSVClinicalTrialInformationExtractor extends Extractor {
         trialSubjectID,
         trialResearchID,
         patientId,
+        trialResearchSystem,
```
julianxcarter, line 45: JC2: If I'm understanding the task correctly, I think we'll still want to keep a default value of `http://example.com/clinicaltrialids` in the case that the trialResearchSystem field is left out of the CSV input. We typically add default values on to optional fields in the extractor, with a good example of this being [this line in the CSVAdverseEventExtractor](https://github.com/mcode/mcode-extraction-framework/blob/9dedb04b7adc9f6c6fb75c44f7354ab8d01597b8/src/extractors/CSVAdverseEventExtractor.js#L34).

test/sample-client-data/clinical-trial-information.csv
```diff
@@ -1,4 +1,4 @@
-mrn,trialSubjectID,enrollmentStatus,trialResearchID,trialStatus
-123,subjectId-1,potential-candidate,researchId-1,approved
-456,subjectId-2,on-study-intervention,researchId-1,completed
-789,subjectId-3,on-study-observation,researchId-2,active
+mrn,trialSubjectID,enrollmentStatus,trialResearchID,trialStatus,trialResearchSystem
+123,subjectId-1,potential-candidate,researchId-1,approved,system-1
+456,subjectId-2,on-study-intervention,researchId-1,completed,system-2
```
julianxcarter, line 3: JC3: If I'm correct in my assumption that we want the extractor to add in a default system if `trialResearchSystem` is left empty, then it'd probably be best to remove the field from one or two of these example rows so that we can cover that scenario in our manual testing.

julianxcarter (changes requested):
JC: This looks really good! I have a few minor comments, mostly surrounding adding in a default system in the case that the `trialResearchSystem` field is left blank. If you have any questions, I'm happy to expand on anything that I said!

test/sample-client-data/clinical-trial-information.csv
dmendelowitz: Committed changes based on all your feedback, give it another look and let me know if you think of anything else

julianxcarter (approved):
This looks good and works well, thanks for making those changes. Nice work! @dmendelowitz

Dtphelan1:
DP1 - One thing I want to point out (shoutout JA and MG for the help) is that these changes should probably result in an update to the E-MEF. Would you be okay making an identically-named branch in that repo and update that repo's ClinicalTrialInformationExtractor to parse off the `trialResearchSystem` in it's `joinAndReformatData` function? There may be some fixtures that need updating as well, but hopefully there won't be a ton of work involved in that branch.

Dtphelan1:
Since that's done I think I'm all set to approve here. 👍

Dtphelan1 (approved):

Dtphelan1:
Sometime tomorrow @dmendelowitz you can rebase and then merge this branch, when you are ready. After your code has been rebased and merged here, you can run `npm update mcode-extraction-framework` from the dependent repo to update the version of the MEF we use to include this newly merged code. After that, you should observe a change to the commit hash in the package-lock.json (importantly: we don't expect a change in package.json). If everything looks good there, you can push the updated commit hash for the MEF. This will allow our CI tests to run appropriately in the dependent repo.

Congrats on your first PR! 🎉 🎉 🎉

PR #88, opened 2021-02-26
Implements the case insensitive CancerDiseaseStatus lookups

Dtphelan1:
# Summary

Implements case insensitive lookups for CancerDiseaseStatus values and introduces a new pattern for where to define  lookup objects, how to export them in a standard lowercase fashion, and what function to use in order to make a case-insensitive queries to lookups.

## New behavior
Introduces case-insensitivity to CancerDiseaseStatus lookup calls by introducing a new `lowercaseLookupQuery` function.

## Code changes
- Replaced the generic `helperUtils.js` with the specific `lookupUtils.js` file;
- Renames `invertObject` to `createInvertedLookup`;
- Creates two new functions `createLowercaseLookup` and `lowercaseLookupQuery`;
- Moves lookups from `diseaseStatusUtils` to a dedicated `helpers/lookup/diseaseStatusLookup.js` file;
- Uses the `createLowercaseLookup` to standardize all lookups before export by the `helpers/lookup/diseaseStatusLookup.js`;
- Adds the `lowercaseLookupQuery` function to all the lookup functions exposed by `diseaseStatusUtils.js`;

# Testing guidance
- Ensure that the new tests pass and provide sufficient coverage of for the new functions;
- Ensure that the old tests pass; 
- Ensure that running the default client still produces CancerDiseaseStatus resources with the appropriate keys and values


julianxcarter (approved):
This looks really good to me, and works well both here and in the epic MEF. Nice work @Dtphelan1!

test/helpers/lookupUtils.test.js
```diff
@@ -0,0 +1,96 @@
...
+      expect(Object.keys(lowercaseLookup)).toEqual(lowercasedExampleKeys);
+    });
+
+    test('values should remain unchanged after lowercasing', () => {
+      const originalValues = Object.values(exampleLookup);
+      expect(Object.values(lowercaseLookup)).toEqual(originalValues);
+    });
+
+    test('# of keys should stay the same', () => {
+      expect(Object.keys(lowercaseLookup).length).toEqual(Object.values(exampleLookup).length);
```
Dtphelan1, line 29:
```suggestion
      expect(Object.keys(lowercaseLookup).length).toEqual(Object.keys(exampleLookup).length);
```

Dtphelan1: Great catch! Thanks Dyl

PR #89, opened 2021-03-02
Fixes the CDS lookup to distinguish between `undetermined` and `not evaluated`

Dtphelan1:
Pretty straightforward. This work fixes a previous oversight in our CDS lookup object, fixing one of our k-v pairs and adding a new one for `not evaluated: not-asked`. I had to make minor adjustments to our CDS-dataAbsentReason logic, so that our Template managed these new. codes appropriately. I've also added a new test to check for dataAbsentReason specific properties.

Addl changes: 
- Changed a few comments in `diseaseStatusUtils.js`, but those are tangentially related to the main task. 

Note: Also looks like more changes were made to `minimal-disease-status-resource.json` then actually were because of linting disparities. C'est la vie. 

**One Assignee should be sufficient**

### DIFF
--- a/test/templates/cancerDiseaseStatus.test.js
+++ b/test/templates/cancerDiseaseStatus.test.js
@@ -48,7 +48,7 @@ describe('test CancerDiseaseStatus template', () => {
       value: {
         code: '709137006',
         system: 'http://snomed.info/sct',
-        display: 'not evaluated',
+        display: 'undetermined',
       },
       subject: {
         id: '123-example-patient',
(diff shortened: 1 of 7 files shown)


test/templates/cancerDiseaseStatus.test.js
```diff
@@ -69,6 +69,38 @@ describe('test CancerDiseaseStatus template', () => {
...
+      },
+      subject: {
+        id: '123-example-patient',
+        name: 'Mr. Patient Example',
+      },
+      condition: {
+        id: '123-Walking-Corpse-Syndrome',
+        name: 'Walking Corpse Syndrome',
+      },
+      effectiveDateTime: '1994-12-09T09:07:00Z',
```
mgramigna, line 88: HBD

test/templates/cancerDiseaseStatus.test.js
```diff
@@ -69,6 +69,38 @@ describe('test CancerDiseaseStatus template', () => {
     expect(isValidFHIR(generatedDiseaseStatus)).toBeTruthy();
   });
 
+  test('valid data where cds-status is not not-asked generates a dataAbsentReason', () => {
```
mgramigna, line 72: MG1: My brain hurts but I think this is an invalid double negative?

Dtphelan1: my b

Dtphelan1: thoughts?

mgramigna:
MG2: Minor, but thoughts on adding a row to the CSV to demonstrate this functionality in practice?

mgramigna (approved):

PR #90, opened 2021-03-08
Fixed hasChanged check within carePlanChangeReasonExtensionTemplate

julianxcarter:
# Summary
The conditional which throws an error if the `hasChanged` element is missing from a review now checks for a `null` value rather than an `undefined` value. 
## New behavior
No new behavior.
## Code changes
1. The conditional within the `carePlanChangeReasonExtensionTemplate` has been updated to check for a `null` value.
2. The missing data test within the carePlanWithReview template test file has been updated to check for the relevant `hasChanged` error explicitly, rather than checking that an error was thrown generally.
# Testing guidance
Ensure that the change to the care plan template tests make sense, as no new behavior should be introduced in this PR.


dmendelowitz (approved):
Everything here looks good. Good work!

jafeltra (approved):
Makes sense to me!

PR #92, opened 2021-03-15
Adding ability to mask fields in Patient resource

dmendelowitz:
# Summary
A set number of fields can now be specified to be masked with dataAbsentReason extensions in a Patient resource

## New behavior
The CSV Patient Extractor and FHIR Patient Extractor can now take an optional constructor argument "mask", which is an array containing any number of fields that are capable of being masked. The specified fields will have their values replaced by dataAbsentReason extensions with the code 'masked'. The fields that can currently be masked are any values the CSV Patient Extractor can take as input (currently "gender", "mrn", "name", "address", "birthDate", "language", "ethnicity", "birthsex" and "race")

## Code changes
- `patientUtils.js` now has a function called `maskPatientData` that takes a FHIR bundle with a Patient resource and an array of fields to mask. maskPatientData will modify the bundle to have dataAbesentReason extensions where necessary 
- `CSVPatientExtractor.js` and `FHIRPatientExtractor.js` now take an optional argument of a 'mask' array and call `maskPatientData` with the mask array as the last step of the extraction process.
- `patientUtils.test.js` now contains tests for `maskPatientData`

# Testing guidance
- Ensure that maskPatientData masks specified fields with proper extensions
- Ensure the output is still a valid FHIR bundle
- Test the functionality on both the MEF and E-MEF
- Check that the new tests pass and test a sufficient number of cases

# Example config:
For mcode-extraction-framework:
`"extractors": [
    {
      "label": "patient",
      "type": "CSVPatientExtractor",
      "constructorArgs": {
        "filePath": "./test/sample-client-data/patient-information.csv",
        "mask": ["gender","mrn","name","address","birthDate","language","ethnicity","birthsex","race"]
      }
    }
  ]`
For epic-mcode-extraction-framework:
`"extractors": [
    {
      "label": "patientExtractor",
      "type": "FHIRPatientExtractor",
      "constructorArgs": {
        "mask": ["gender","mrn","name","address","birthDate","language","ethnicity","birthsex","race"]
      }
    }
  ]`
(In both configs `mask` contains all possible fields, feel free to try any combination)

### DIFF
--- a/src/helpers/patientUtils.js
+++ b/src/helpers/patientUtils.js
@@ -91,8 +91,7 @@ function maskPatientData(bundle, mask) {
 
   mask.forEach((field) => {
     if (!validFields.includes(field)) {
-      throw Error(`'${field}' is not a field that can be masked. Patient will only be extracted if all mask fields are valid. \
-      Valid fields include: 'gender','mrn','name','address','birthDate','language','ethnicity','birthsex','race'`);
+      throw Error(`'${field}' is not a field that can be masked. Patient will only be extracted if all mask fields are valid. Valid fields include: Valid fields include: ${validFields.join(', ')}`);
     }
     // must check if the field exists in the patient resource, so we don't add unnecessary dataAbsent extensions
     if (field === 'gender' && 'gender' in patient) {
--- a/src/helpers/patientUtils.js
+++ b/src/helpers/patientUtils.js
@@ -69,7 +69,7 @@ function getRaceDisplay(code) {
  * @return {string} concatenated string of name values
  */
 function getPatientName(name) {
-  return `${name[0].given.join(' ')} ${name[0].family}`;
+  return ('extension' in name[0]) ? 'masked' : `${name[0].given.join(' ')} ${name[0].family}`;
 }
 
 /**
--- a/src/helpers/patientUtils.js
+++ b/src/helpers/patientUtils.js
@@ -1,3 +1,7 @@
+/* eslint-disable no-underscore-dangle */
+const fhirpath = require('fhirpath');
+const { extensionArr, dataAbsentReasonExtension } = require('../templates/snippets/extension.js');
+
 // Based on the OMB Ethnicity table found here:http://hl7.org/fhir/us/core/STU3.1/ValueSet-omb-ethnicity-category.html
 const ethnicityCodeToDisplay = {
   '2135-2': 'Hispanic or Latino',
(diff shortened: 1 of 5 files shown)


dmendelowitz:
> This PR looks really good to me! I had two small suggestions that I've made a note of.
> 
> I tested this with the Epic-MEF as well and it worked great for the most part. ~I do think we'll need a smaller PR over there to handle cases within the [EpicCancerDiseaseStatusExtractor](https://gitlab.mitre.org/mcode/epic-mcode-extraction-framework/-/blob/master/src/extractors/EpicCancerDiseaseStatusExtractor.js#L78) and [EpicTreatmentPlanChangeExtractor](https://gitlab.mitre.org/mcode/epic-mcode-extraction-framework/-/blob/master/src/extractors/EpicTreatmentPlanChangeExtractor.js#L162) where they could potentially try to pull the patient's name for those resources after a user has specified it to be masked. I'm not sure if that's in scope for this task, but just something to consider.~
> 
> Scratch that, I actually think we could make this change within the [getPatientName function in the MEF](https://github.com/mcode/mcode-extraction-framework/blob/1bbbbe40b18efd7c3b6865f4df1a05fa406831d7/src/helpers/patientUtils.js#L68)? We should definitely discuss if that function should return a `Masked for privacy reasons` value or something like that.

Is the getPatientName function actually used anywhere in the MEF? I only see it in the patientUtils file and then in the test file. Knowing the context would definitely help with figuring out how it should act when the field is masked.

src/helpers/patientUtils.js
```diff
@@ -65,12 +69,88 @@ function getRaceDisplay(code) {
  * @return {string} concatenated string of name values
  */
 function getPatientName(name) {
-  return `${name[0].given.join(' ')} ${name[0].family}`;
+  return ('extension' in name[0]) ? 'masked' : `${name[0].given.join(' ')} ${name[0].family}`;
```
dmendelowitz, line 72: Looking for opinions on what should actually be returned here when the name field is masked (this is used in EpicCancerDiseaseStatus and EpicTreatmentPlanChange Extractors for context)

jafeltra: I think the string `'masked'` is good! I don't think it's safe to return null or undefined since it might get applied to a required field, and I think it might be too much for right now to support returning a full dataAbsent extesion.

jafeltra (commented):
This looks really good! I also just had one small comment, but this is a really cool feature!

dmendelowitz:
I think all comments have been addressed now, give it another look and let me know if anything else comes up

src/helpers/patientUtils.js
```diff
@@ -91,7 +91,8 @@ function maskPatientData(bundle, mask) {
 
   mask.forEach((field) => {
     if (!validFields.includes(field)) {
-      throw Error(`'${field}' is not a field that can be masked. Valid fields include: 'gender','mrn','name','address','birthDate','language','ethnicity','birthsex','race'`);
+      throw Error(`'${field}' is not a field that can be masked. Patient will only be extracted if all mask fields are valid. \
+      Valid fields include: 'gender','mrn','name','address','birthDate','language','ethnicity','birthsex','race'`);
```
jafeltra, line 95: I really like the clarification here and I think it will be helpful! Did you want the Valid fields list to be on the next line? Right now, it stays on the same line with a bunch of extra spaces before "Valid" for me, which ideally wouldn't be there. I think the extra spaces are from the indent in the code since it's all within the same string.

dmendelowitz: I can fix that, I just put the slash there because lint didn't like a really long line in the code

jafeltra: Oh rightt. Hmm because it still won't like the long line and we probably shouldn't just disable it. Maybe we put the error message into a variable and just use string concatenation? I feel like we've hit something similar for another long error message, but I can't find it now.

jafeltra (approved):
This looks awesome!

julianxcarter (approved):
This looks great to me!

PR #93, opened 2021-03-16
Example config rename

Dtphelan1:
Renamed the example config file and updated documentation/gitignores. Should only need one reviewer.


dmendelowitz (approved):
Looks good!

PR #95, opened 2021-03-19
Removed default value for trialResearchSystem in ClinicalTrialInformationExtractor

dmendelowitz:
# Summary
trialResearchSystem will no longer default to an "example.com" value when not provided in the CSV
## New behavior
When no trialResearchSystem is provided in the CSV, the 'system' field will be omitted instead of given a default value
## Code changes
CSVClinicalTrialInformationExtractor.js no longer assigns an 'example.com' default value to trialResearchSystem
# Testing guidance
- Make sure that when no trialResearchSystem is provided, the resulting ResearchStudy has an identifier with no 'system' field
- Make sure the extractor still works as expected otherwise and tests still pass

### DIFF
--- a/src/extractors/CSVClinicalTrialInformationExtractor.js
+++ b/src/extractors/CSVClinicalTrialInformationExtractor.js
@@ -27,8 +27,6 @@ class CSVClinicalTrialInformationExtractor extends BaseCSVExtractor {
     const {
       trialSubjectID, enrollmentStatus, trialResearchID, trialStatus, trialResearchSystem,
     } = clinicalTrialData;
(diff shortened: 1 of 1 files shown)


Dtphelan1:
One small thing: we can fully remove the default value line instead of commenting it.

dmendelowitz:
> One small thing: we can fully remove the default value line instead of commenting it.

Oops, forgot to delete that

Dtphelan1:
This looks good otherwise!

Dtphelan1 (approved):

julianxcarter (approved):

PR #97, opened 2021-03-23
Create default log file if one does not exist

dmendelowitz:
# Summary
If a log file does not exist and no custom log file is specified, a new file is created at the default location.
## New behavior
When running the program with the `--entries-filter` flag, if no log file exists at `logs/run-logs.json`, then the file is created and initialized with an empty array. This won't happen if a custom log file is specified.
## Code changes
- `checkLogFile` function in `app.js` now creates a default log file if one doesn't exist.
- `README.md` has been updated to better reflect the logging functionality
# Testing guidance
- Test that a log file is created when there is no file at `logs/run-logs.json`
- Example command to run: `node src/cli/cli.js -e -f 2019-01-01 -t 2021-01-01`
- Test that specifying a custom log file still works as it should
- Make sure the changes to the README accurately reflect how logging works

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ If the `notificationInfo` object is provided in configuration, an email will be 
 
 ### Logging Successful Extractions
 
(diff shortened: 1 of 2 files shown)


mgramigna (approved):

jafeltra (approved):
This looks great and the README is much clearer now!

PR #98, opened 2021-03-26
Mask MRN when used as the ID of a Patient resource

dmendelowitz:
# Summary
Since the MRN is used as an ID on the patient resource, it appears as a reference in other resources. Now when masking the MRN, all of those references will also be masked by replacing the MRN with a hash of the MRN after extraction is completed. 
## New behavior
If the MRN is specified as a field to be masked, in addition to masking the MRN in the Patient resource, there is an additional masking step after extraction is finished. All references to the MRN as an ID will be replaced by a hashed version of the MRN. Right now, the only places I found the MRN being used as an ID are in the `id` and `fullUrl` fields of the Patient resource and 'subject' and 'individual' objects in other resources.
## Code changes
- `patientUtils.js` now has a `maskMRN` function that takes a bundle as input, modifies the bundle to replace the MRN with a hashed version as the ID of the Patient resource, and then replaces every reference in other resources with the same hash 
- `app.js` will check the config to see if the MRN needs to be masked and will perform the masking after extraction is completed
- `patientUtils.test.js` now contains tests for the `maskMRN` function
# Testing guidance
- Run the program with `"mask": ["mrn"]` in the `constructorArgs` of the patient extractor
- Make sure that when MRN is specified to be masked, there are now no references to the MRN anywhere in the output bundle
- Check to see if the MRN is used in any locations that I missed
- Make sure all tests pass and the program runs as expected

### DIFF
--- a/src/cli/app.js
+++ b/src/cli/app.js
@@ -113,7 +113,7 @@ async function mcodeApp(Client, fromDate, toDate, pathToConfig, pathToRunLogs, d
     // check if config specifies that MRN needs to be masked
     // if it does need to be masked, mask all references to MRN outside of the patient resource
     const patientConfig = config.extractors.find((e) => e.type === 'CSVPatientExtractor');
-    if ('constructorArgs' in patientConfig && 'mask' in patientConfig.constructorArgs) {
+    if (patientConfig && ('constructorArgs' in patientConfig && 'mask' in patientConfig.constructorArgs)) {
       if (patientConfig.constructorArgs.mask.includes('mrn')) {
         extractedData.forEach((bundle) => {
           maskMRN(bundle);
--- a/src/cli/app.js
+++ b/src/cli/app.js
@@ -6,6 +6,7 @@ const logger = require('../helpers/logger');
 const { RunInstanceLogger } = require('./RunInstanceLogger');
 const { sendEmailNotification, zipErrors } = require('./emailNotifications');
 const { extractDataForPatients } = require('./mcodeExtraction');
+const { maskMRN } = require('../helpers/patientUtils');
 
 function getConfig(pathToConfig) {
   // Checks pathToConfig points to valid JSON file
@@ -106,6 +107,17 @@ async function mcodeApp(Client, fromDate, toDate, pathToConfig, pathToRunLogs, d
       const successCondition = successfulExtraction;
       if (successCondition) {
         runLogger.addRun(effectiveFromDate, effectiveToDate);
+      }
+    }
+
+    // check if config specifies that MRN needs to be masked
+    // if it does need to be masked, mask all references to MRN outside of the patient resource
+    const patientConfig = config.extractors.filter((e) => e.type === 'CSVPatientExtractor')[0];
+    if ('constructorArgs' in patientConfig && 'mask' in patientConfig.constructorArgs) {
+      if (patientConfig.constructorArgs.mask.includes('mrn')) {
+        extractedData.forEach((bundle) => {
+          maskMRN(bundle);
+        });
       }
     }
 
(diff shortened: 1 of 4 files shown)


src/cli/app.js
```diff
@@ -109,6 +110,17 @@ async function mcodeApp(Client, fromDate, toDate, pathToConfig, pathToRunLogs, d
       }
     }
 
+    // check if config specifies that MRN needs to be masked
+    // if it does need to be masked, mask all references to MRN outside of the patient resource
+    const patientConfig = config.extractors.filter((e) => e.type === 'CSVPatientExtractor')[0];
```
mgramigna, line 115:
```suggestion
    const patientConfig = config.extractors.find((e) => e.type === 'CSVPatientExtractor');
```

`find` instead of `filter`[0] probably makes sense here. `find` will return the first occurrence that matches the boolean you provide, and `undefined` if not found

src/cli/app.js
```diff
@@ -109,6 +110,17 @@ async function mcodeApp(Client, fromDate, toDate, pathToConfig, pathToRunLogs, d
...
+    if ('constructorArgs' in patientConfig && 'mask' in patientConfig.constructorArgs) {
```
mgramigna, line 116:
```suggestion
    if (patientConfig && ('constructorArgs' in patientConfig && 'mask' in patientConfig.constructorArgs)) {
```

In tandem with above `find` suggestion, might as well guard against the undefined case here. Even though I think the `if` will still work find without it. It's just a little bit more expressive

src/helpers/patientUtils.js
```diff
@@ -147,10 +148,33 @@ function maskPatientData(bundle, mask) {
...
 }
 
+/**
+ * Mask all references to the MRN used as an id
+ * Currently, the MRN appears as an id in 'subject' and 'individual' objects in other resources
+ * and in the 'id' and 'fullUrl' fields of the Patient resource.
+ * Replaces the MRN with a hash of the MRN
+ * @param {Object} bundle a FHIR bundle with a Patient resource and other resources
+ */
+function maskMRN(bundle) {
```
mgramigna, line 158: Nice use of fhirpath here

src/helpers/patientUtils.js
```diff
@@ -147,10 +148,33 @@ function maskPatientData(bundle, mask) {
...
+  const patient = fhirpath.evaluate(bundle, 'Bundle.entry.where(resource.resourceType=\'Patient\')')[0];
+  if (patient === undefined) throw Error('No Patient resource in bundle. Could not mask MRN.');
+  const mrn = patient.resource.id;
+  const masked = shajs('sha256').update(mrn).digest('hex');
+  patient.fullUrl = `urn:uuid:${masked}`;
+  patient.resource.id = masked;
+  const subjects = fhirpath.evaluate(bundle, `Bundle.entry.resource.subject.where(reference='urn:uuid:${mrn}')`);
+  const individuals = fhirpath.evaluate(bundle, `Bundle.entry.resource.individual.where(reference='urn:uuid:${mrn}')`);
+  const mrnOccurances = subjects.concat(individuals);
```
mgramigna, line 167:
```suggestion
  const mrnOccurrences = subjects.concat(individuals);
``` 

Heheh sorry. Would also have to update the uses of it below

src/helpers/patientUtils.js
```diff
@@ -147,10 +148,33 @@ function maskPatientData(bundle, mask) {
...
+  for (let i = 0; i < mrnOccurances.length; i += 1) {
+    mrnOccurances[i].reference = `urn:uuid:${masked}`;
+  }
```
mgramigna, line 170: Clarifying Q: I'm assuming this is a for .. let instead of a forEach because forEach will yell at you for no function param re-assign, right?

mgramigna (changes requested):
Overall the masking changes work/look great! I just have a few small comments and suggestions that I made inline

src/helpers/patientUtils.js
dmendelowitz: Yep, forEach seemed cleaner to me, but I assumed there was a reason for the lint rule, so I went with the for loop

mgramigna (approved):
Looks great!

julianxcarter (approved):
Looks good to me and works well. Nice job!

PR #99, opened 2021-03-29
Support specifying clinicalSiteSystem in configuration file

dmendelowitz:
# Summary
clinicalSiteSystem can now be specified in the config file, similar to how clinicalSiteID is, except clinicalSiteSystem is optional
## New behavior
If a `clinicalSiteSystem` is added to the `constructorArgs` of `CSVClinicalTrialInformationExtractor` in the config file, that value will be used for the `system` field of the site object in the ResearchStudy resource. If no system is specified, the `system` field will be omitted in the output.
## Code changes
- `CSVClinicalTrialInformationExtractor.js` now takes an optional `clinicalSiteSystem` constructor argument and adds that to the formatted data for the Research Study
- `ResearchStudyTemplate.js` now takes an input for `clinicalSiteSystem` instead of just using an example.com value
- `CSVClinicalTrialInformationExtractor.test.js` and `researchStudy.test.js` have been updated to reflect the above changes
# Testing guidance
- Add a value for `clinicalSiteSystem` to the constructor arguments of the `CSVClinicalTrialInformationExtractor` and see that the correct value appears in the output and that there is no system in the output if the argument is removed from the config
- Make sure the new tests pass and accurately test the new functionality

### DIFF
--- a/src/extractors/CSVClinicalTrialInformationExtractor.js
+++ b/src/extractors/CSVClinicalTrialInformationExtractor.js
@@ -16,10 +16,11 @@ function getPatientId(context) {
 }
 
 class CSVClinicalTrialInformationExtractor extends BaseCSVExtractor {
-  constructor({ filePath, clinicalSiteID }) {
+  constructor({ filePath, clinicalSiteID, clinicalSiteSystem }) {
     super({ filePath, csvSchema: CSVClinicalTrialInformationSchema });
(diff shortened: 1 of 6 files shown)


julianxcarter (approved):
Looks good to me!

mgramigna (approved):

PR #102, opened 2021-04-01
Add Treatment Plan Change CSV

jafeltra:
# Summary
This PR adds a Treatment Plan Change sheet to the Excel document. When looking through the example CSVs we have, I noticed that it looks like we support a `reasonDisplayText` in extraction that wasn't in our CSVs, so I added that. When reviewing this PR:
- Ensure that the Excel sheet changes make sense and reflect the state of the extractor
- I tested out the reasonDisplayText and it works as I'd expect, but I'm also not sure how we included it in the extractor and not the CSVs, so make sure I'm not missing something and we don't want that column
    - If this looks correct, I can make an equivalent PR for that change in the Base ICARE client repo
- There's technically a `dateRecorded` column that we support in the CSV, but since it is used for entry filtering based on from and to dates and it's use is not encouraged, I didn't include it in Excel. This seems to be the same pattern that CancerDiseaseStatus follows as well.


Dtphelan1 (approved):
This looks great!

PR #106, opened 2021-04-14
Adding npm start script

dmendelowitz:
# Summary
You can now call `npm start -- [options]` to start the client instead of `node src/cli/cli.js [options]`
## New behavior
There is now a script in `package.json` called "start" that acts as a simpler way of calling the program from the command line. It is important to note that npm requires you to put "--" between your script name and command line arguments. If the double hyphen is left out, the arguments will be interpreted as npm flags, for example -d and -f will be interpreted as npm's debug and force flags, not our debug and from date flags.
## Code changes
- `package.json` now has a start script, which runs the client
- The README has been updated to include the new way of running the client
# Testing guidance
Make sure that you are able to use the client as usual when calling via the start script

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,19 +58,13 @@ const cancerDiseaseStatusExtractor = new CSVCancerDiseaseStatusExtractor('path-t
 The framework also contains the [MCODEClient](src/client/MCODEClient.js) which has registered all of the extractors in this repo. Once you have exported CSV data and updated your configuration file, use the mCODE Extraction client by running the following:
 
 ```bash
-node src/cli/cli.js [options]
+npm start -- [options]
 ```
 
 To see all the options that can be used with the mCODE client, run the following:
 
 ```bash
-node src/cli/cli.js --help
-```
-
-The client can also be run by npm script:
-
-```bash
-npm start -- [options]
+npm start -- --help
 ```
 
 ### First Time User Guide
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -67,6 +67,12 @@ To see all the options that can be used with the mCODE client, run the following
 node src/cli/cli.js --help
 ```
 
+The client can also be run by npm script:
```
julianxcarter, line 70: I wonder if it'd be a good idea to entirely remove the instructions in the readme for running the client by way of the `node src/cli/cli.js` command. It seems like something that could only introduce confusion to some of our users, given that the node command only works from the top-level directory but the `npm start` command works from anywhere in the project. I'd love to hear your and @Dtphelan1's thoughts on this . 

If we do keep the instructions for running the client with the `node src/cli/cli.js` command, we may want to move the `npm start` instructions to be the first command described in this readme and kind of suggest to users that this is our recommended way for running the client.

package.json
```diff
@@ -12,6 +12,7 @@
   ],
   "main": "src/",
   "scripts": {
+    "start":"node src/cli/cli.js",
```
julianxcarter, line 15: I think it'd also be a good idea to add a `help` script here that corresponds to `node src/cli/cli.js --help` so that users can have a script to see our cli options as well.

julianxcarter (commented):
This looks great and works really well for me! I have two small suggestions that I'd like to hear your thoughts on.

README.md
dmendelowitz: Yeah, I wasn't sure if this was meant to be a replacement to `node src/cli/cli.js` or just an extra option. If we want to push this as the main way to run the client I can definitely make changes to the README to reflect that. I would like to hear other's opinions as well though.

package.json
dmendelowitz: The user could always run `npm start -- --help`, but do you mean it might be helpful to have something like `npm help` that is an alias for `node src/cli/cli.js --help`?

dmendelowitz: Actually `npm help` already a built in command, so we'd have to call it something else

julianxcarter: Ahhh yes I missed what you said earlier about the `--` being needed before any of our cli options. I had run `npm start --help` and it didn't work for me before. Nevermind then!

julianxcarter (approved):
Looks good to me!

PR #107, opened 2021-04-14
CSV Module now ignores BOMs

julianxcarter:
# Summary
This PR enables the CSV Module to read csvs with invisible Byte Order Marks at the beginning of a file.
## New behavior
The extractor now runs as expected when csvs containing a client's data include Byte Order Marks at the beginning of the file.
## Code changes
1. The `parse` function that's run in the CSVModule's contractor now has the `bom` flag set to true, causing any BOMs at the beginning of the csv to be ignored.
2. A new `example-csv-bom.csv` file has been added as a test fixture. It's identical to the preexisting `example-csv.csv` file, except it include a BOM at the beginning of the file.
3. A test case verifying that the CSV module ignores BOMs has been added to `CSVModule.test.js`
# Testing guidance
Commit `3dd1055` adds both an example csv with a BOM as well as a unit test that attempts to parse this CSV. The `bom` flag has __not__ been added in this commit though, so you should see this unit test fail as the CSVModule fails to recognize the `mrn` column. In commit `0e3d63a`, that same unit test should pass as that commit enables the `bom` flag.

### DIFF
--- a/src/modules/CSVModule.js
+++ b/src/modules/CSVModule.js
@@ -5,7 +5,7 @@ const logger = require('../helpers/logger');
 
 class CSVModule {
   constructor(csvFilePath) {
-    this.data = parse(fs.readFileSync(csvFilePath), { columns: true });
+    this.data = parse(fs.readFileSync(csvFilePath), { columns: true, bom: true });
   }
 
   async get(key, value, fromDate, toDate) {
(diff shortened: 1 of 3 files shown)


dmendelowitz (approved):
This looks good to me!

mgramigna (approved):

PR #108, opened 2021-04-16
Numerous Extractor Updates

Dtphelan1:
# Summary
~We don't mention scope creep...~ So this PR looks like it does a lot more than it actually does. Originally, the goal of this PR covered a few items: 
- When one doesn't get data back from the Module, the extractor should communicate that information and return an empty bundle. 
- We shouldn't be using the MRN as our patientId now that we have Context – let's use the ID off of the existing patient resource. 

This second change very gently suggested two other reasonable (if not sprawling) changes: 
- If the subjectId isn't an MRN, we probably shouldn't be passing that information into templates using the MRN field. 
- If the patient's ID is parsed off context, we shouldn't enforce a requirement on the `MRN` property in joinAndReformat. The MRN requirement should be a result of the CSVmodule.get, not a result of the reformatData function.

Overall, these are very small changes that stretch wide distances. Additionally, since changes take place at such a foundational level (e.g. in some cases the interfaces for generatingMcodeResources has changed) there is a sibling PR in the E-MEF. 

## Testing: 
- Ensure that all tests work locally 
- As best as you can, review the following on CSVExtractors: 
    - Look and feel of the `get` routine
    - Any leftover uses of the `mrn` beyond the original CSVModule call 
    - The validity of data passed into the Templates.
- Ensure that the cli works and produces data that looks right.  
- Anything else you and think of!

### DIFF
--- a/src/extractors/CSVCancerRelatedMedicationExtractor.js
+++ b/src/extractors/CSVCancerRelatedMedicationExtractor.js
@@ -15,7 +15,7 @@ function formatData(medicationData, patientId) {
     } = medication;
 
     if (!(code && codeSystem && status)) {
-      throw new Error('The cancer-related medication is missing an expected element; mrn, code, code system, and status are all required values.');
+      throw new Error('The cancer-related medication is missing an expected element; code, code system, and status are all required values.');
     }
 
     return {
--- a/src/extractors/CSVCancerRelatedMedicationExtractor.js
+++ b/src/extractors/CSVCancerRelatedMedicationExtractor.js
@@ -1,24 +1,26 @@
 const { BaseCSVExtractor } = require('./BaseCSVExtractor');
 const { generateMcodeResources } = require('../templates');
-const logger = require('../helpers/logger');
+const { getPatientFromContext } = require('../helpers/contextUtils');
+const { getEmptyBundle } = require('../helpers/fhirUtils');
 const { formatDateTime } = require('../helpers/dateUtils');
+const logger = require('../helpers/logger');
 
 
-function formatData(medicationData) {
+function formatData(medicationData, patientId) {
   logger.debug('Reformatting cancer-related medication data from CSV into template format');
 
   return medicationData.map((medication) => {
     const {
-      mrn, medicationId, code, codeSystem, displayText, startDate, endDate, treatmentReasonCode, treatmentReasonCodeSystem, treatmentReasonDisplayText, treatmentIntent, status,
+      medicationId, code, codeSystem, displayText, startDate, endDate, treatmentReasonCode, treatmentReasonCodeSystem, treatmentReasonDisplayText, treatmentIntent, status,
     } = medication;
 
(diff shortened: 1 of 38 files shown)


test/extractors/CSVCancerRelatedMedicationExtractor.test.js
```diff
@@ -35,24 +37,25 @@ describe('CSVCancerRelatedMedicationExtractor', () => {
     test('should join data appropriately and throw errors when missing required properties', () => {
       const expectedErrorString = 'The cancer-related medication is missing an expected element; mrn, code, code system, and status are all required values.';
       const localData = _.cloneDeep(exampleCSVMedicationModuleResponse);
+      const patientId = getPatientFromContext(MOCK_CONTEXT).id;
 
       // Test that valid maximal data works fine
-      expect(formatData(exampleCSVMedicationModuleResponse)).toEqual(expect.anything());
+      expect(formatData(localData, patientId)).toEqual(expect.anything());
```
jafeltra, line 43: Oh good catch on these `localData`s!

jafeltra (commented):
This looks great to me! I ran it with all the extractors in my config and all the resulting bundles are equivalent to what is on `develop`, and those few errors about destructuring are resolved. I had one incredibly minor comment in line and just one questions about tests:

Do you think a test that each extractor returns an empty bundle if there are no entries returned from the CSV module would be worthwhile? I'm not sure if it's overkill to test that, but it's the only new functionality that I could think of that isn't already tested.

src/extractors/CSVCancerRelatedMedicationExtractor.js
```diff
@@ -1,24 +1,26 @@
...
 
   return medicationData.map((medication) => {
     const {
-      mrn, medicationId, code, codeSystem, displayText, startDate, endDate, treatmentReasonCode, treatmentReasonCodeSystem, treatmentReasonDisplayText, treatmentIntent, status,
+      medicationId, code, codeSystem, displayText, startDate, endDate, treatmentReasonCode, treatmentReasonCodeSystem, treatmentReasonDisplayText, treatmentIntent, status,
     } = medication;
 
-    if (!(mrn && code && codeSystem && status)) {
+    if (!(code && codeSystem && status)) {
       throw new Error('The cancer-related medication is missing an expected element; mrn, code, code system, and status are all required values.');
```
Dtphelan1, line 18:
```suggestion
      throw new Error('The cancer-related medication is missing an expected element; code, code system, and status are all required values.');
```

jafeltra (approved):
I made a [Jira task](https://jira.mitre.org/browse/STEAM-532) for the empty bundle tests. This looks great!

mgramigna (changes requested):
Can you update the example csv config JSON file to have the patient extractor first so our example would be kosher with context in a hypothetical extraction situation?

mgramigna (approved):
Looks good!

PR #110, opened 2021-04-21
Make CSV column headers case insensitive

dmendelowitz:
# Summary
Column headers of CSV files read as input for the MEF can now have any casing and will still be treated the same by the program 
## New behavior
The program should still run the same as before, but now CSV headers can have any casing (e.g. `enrollmentStatus`, `ENROLLMENTSTATUS`, and `EnRoLlMeNtStAtUs` are all valid and refer to the same thing).
## Code changes
- `csvModule.js` now converts all headers to lowercase upon reading the CSV and will convert any keys provided to a get function to lowercase to properly find values
- Every single CSV extractor has had its version of a formatting function changed to accommodate lowercase csv headers
- Every CSV extractor test file has had its fixtures and tests changed to expect lowercase headers from the csvModule
- `csvValidoator.js` has been updated to convert the CSV schema and actual CSV headers to lowercase before comparing them
# Testing guidance
- Ensure that all previous functionality still works as expected
- Mess up a few column headers with different casing and see that the extractor still works
- Check that any change to some part of the program to accommodate case insensitivity hasn't messed up something else in some unforeseen way

### DIFF
--- a/test/extractors/CSVCancerDiseaseStatusExtractor.test.js
+++ b/test/extractors/CSVCancerDiseaseStatusExtractor.test.js
@@ -35,7 +35,7 @@ describe('CSVCancerDiseaseStatusExtractor', () => {
       expect(csvCancerDiseaseStatusExtractor.joinAndReformatData(exampleCSVDiseaseStatusModuleResponse, patientId)).toEqual(expect.anything());
 
       localData[0].evidence = ''; // Evidence is not required and will not throw an error
-      localData[0].observationStatus = ''; // Observation Status is not required and will not throw an error
+      localData[0].observationstatus = ''; // Observation Status is not required and will not throw an error
 
       // Only including required properties is valid
       expect(csvCancerDiseaseStatusExtractor.joinAndReformatData(localData, patientId)).toEqual(expect.anything());
--- a/test/extractors/CSVCancerDiseaseStatusExtractor.test.js
+++ b/test/extractors/CSVCancerDiseaseStatusExtractor.test.js
@@ -40,7 +40,7 @@ describe('CSVCancerDiseaseStatusExtractor', () => {
       // Only including required properties is valid
       expect(csvCancerDiseaseStatusExtractor.joinAndReformatData(localData, patientId)).toEqual(expect.anything());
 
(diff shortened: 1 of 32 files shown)


test/extractors/CSVProcedureExtractor.test.js
```diff
@@ -41,7 +41,7 @@ describe('CSVProcedureExtractor', () => {
       expect(formatData(localData, patientId)).toEqual(expect.anything());
```
jafeltra, line 41: GitHub won't let me comment on the line I want, but do you know why `bodySite` doesn't need to be updated to `bodysite`? Maybe it doesn't fail because it's optional, but I think we should update it for consistency.

test/extractors/CSVCancerDiseaseStatusExtractor.test.js
```diff
@@ -40,7 +40,7 @@ describe('CSVCancerDiseaseStatusExtractor', () => {
       // Only including required properties is valid
```
jafeltra, line 40: GitHub won't let me comment on the right line, but do you think `observationStatus` should be `observationstatus`?

jafeltra (commented):
This looks great! The changes are really minimal given that you had to change so much! I had a few small things inline and then I had a few other thoughts:

- It looks like [this line in CSVObservationExtractor.test.js](https://github.com/mcode/mcode-extraction-framework/blob/main/test/extractors/CSVObservationExtractor.test.js#L38) should also be updated if you agree with the other similar comments.
- It might be good to test that the CSV Validator can successfully validate headers that are in various casing that doesn't match the schema. I think that should cover the changes in `csvValidator.js` and `CSVModule.js`.
- I think we'll need a small PR in the E-MEF because the ClinicalTrialExtractor there uses the CSV Module, so when we update the MEF, I think it will be looking for the lower case changes.

julianxcarter (approved):
This looks great!

jafeltra (approved):
This looks great! Thanks for making those changes!

jafeltra:
I think this just needs a rebase to get back in line with `develop`, and then it should be good to merge!

PR #111, opened 2021-04-21
Adding references to FHIR Resources in CSV_Templates

dmendelowitz:
# Summary
Each sheet in the CSV_Templates excel file now links to pages with information on relevant FHIR resources to that csv template
# Testing guidance
Check that the links on each sheet are the correct and make sure I didn't leave any out. Also let me know if you have issues with the formatting, I based it on what was already there in the Adverse Event sheet.


jafeltra:
This looks great! I think the one link I'd recommend updating is the one in CancerPatient. The other links use the released STU1 mCODE links, rather than the `build.fhir.org` links, which update as the mCODE repo updates before balloting. So the CancerPatient link should be http://hl7.org/fhir/us/mcode/StructureDefinition-mcode-cancer-patient.html. All the others look correct!

julianxcarter (commented):
All of these changes look good to me, apart from the CancerPatient sheet that @jafeltra referenced. We should also probably delete that `~$CSV_Templates.xlsx` file that was saved as well. Every time you make a change to the file excel will auto-save that to the repo, so we'll need to manually delete it before the next commit.

julianxcarter (approved):
Looks good to me!

jafeltra (approved):
This looks great!

PR #112, opened 2021-04-23
Adding troubleshooting guide for BOMs

dmendelowitz:
# Summary
Added section to README informing users that byte order markers are handled for UTF-8 and UTF-16 but may cause issues in other encodings and how they can check for BOMs in their CSVs
# Testing guidance
Read the new section of the README and make sure it is all correct and that I didn't leave anything out

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ The extraction client has built-in handling of byte order markers for CSV files 
 cat -v <file.csv>
 ```
 
-If there is an unexpected symbol at the beginning of the file, then there is a byte order marker that needs to be removed.
+If there is an unexpected symbol at the beginning of the file, then there may be a byte order marker that needs to be removed.
 
 ## Terminology and Architecture
 
--- a/README.md
+++ b/README.md
@@ -167,10 +167,10 @@ npm start -- --entries-filter --from-date <YYYY-MM-DD> --to-date <YYYY-MM-DD> --
 
 #### Byte Order Markers in CSV Files
 
-The extraction client has built-in handling of byte order markers for CSV files in UTF-8 and UTF-16 encodings. When using CSV files in other encodings, if you experience unexpected errors be sure to check for a byte order marker at the beginning of the file. One way to check is to run the following command from the command line:
+The extraction client has built-in handling of byte order markers for CSV files in UTF-8 and UTF-16LE encodings. When using CSV files in other encodings, if you experience unexpected errors be sure to check for a byte order marker at the beginning of the file. One way to check is to run the following command from the command line:
 
 ```bash
-cat -v file.csv
+cat -v <file.csv>
 ```
 
 If there is an unexpected symbol at the beginning of the file, then there is a byte order marker that needs to be removed.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -161,6 +163,18 @@ If a `from-date` is provided as an option when running the mCODE Extraction Clie
...
 ```
 
+### Troubleshooting
+
+#### Byte Order Markers in CSV Files
+
+The extraction client has built-in handling of byte order markers for CSV files in UTF-8 and UTF-16 encodings. When using CSV files in other encodings, if you experience unexpected errors be sure to check for a byte order marker at the beginning of the file. One way to check is to run the following command from the command line:
+
+```bash
+cat -v file.csv
```
julianxcarter, line 173: I think we should write the example code as `cat -v <path-to-csv>` or `cat -v <file.csv>`. We've been including example values within the `<` and `>` brackets as a general practice in our READMEs.

README.md
```diff
@@ -161,6 +163,18 @@ If a `from-date` is provided as an option when running the mCODE Extraction Clie
 npm start -- --entries-filter --from-date <YYYY-MM-DD> --to-date <YYYY-MM-DD> --config-filepath <path>
 ```
 
+### Troubleshooting
+
+#### Byte Order Markers in CSV Files
+
+The extraction client has built-in handling of byte order markers for CSV files in UTF-8 and UTF-16 encodings. When using CSV files in other encodings, if you experience unexpected errors be sure to check for a byte order marker at the beginning of the file. One way to check is to run the following command from the command line:
```
julianxcarter, line 170: We should specify that we only handle UTF-16 (LE) encodings. There's actually two UTF-16 encodings (LE and BE) and the module only parses out BOMs from the LE variety.

julianxcarter (changes requested):
The here text looks good! I just had two minor comments.

julianxcarter (approved):
Looks great, thanks for handling that!

README.md
```diff
@@ -161,6 +163,18 @@ If a `from-date` is provided as an option when running the mCODE Extraction Clie
...
+
+#### Byte Order Markers in CSV Files
+
+The extraction client has built-in handling of byte order markers for CSV files in UTF-8 and UTF-16LE encodings. When using CSV files in other encodings, if you experience unexpected errors be sure to check for a byte order marker at the beginning of the file. One way to check is to run the following command from the command line:
+
+```bash
+cat -v <file.csv>
+```
+
+If there is an unexpected symbol at the beginning of the file, then there is a byte order marker that needs to be removed.
```
jafeltra, line 176: This is a little nit-picky, but can we maybe update this to say "...there may be a byte order marker that needs to be removed"? I'm guessing there could be a case where the unexpected symbol isn't necessarily a byte order marker or shouldn't be be removed.

jafeltra (commented):
I just had one really small comment. Let me know what you think.

jafeltra (approved):
Looks great!

jafeltra:
I think this might just need to be rebased again. I doubt there are any conflicts but I just want to make sure no README updates get lost!

PR #114, opened 2021-05-06
update docs to have correct construction of extractor

mgramigna:
I noticed this when I was making the demo electron app to construct and call an extractor. This is out of date since the args are objects now


jafeltra (approved):
Amazing

PR #119, opened 2021-05-18
Null is empty csv

Dtphelan1:
# Summary

Adds behavior to the CSVModule to support normalization of NULL/NIL values into empty-string values. Additionally, CSVExtractors can now specify unalterableColumns in order to leave certain types of data immune from this normalization. 

## New behavior
- CSVModule now preprocesses CSV data in order to replace any values where `value.toLowerCase() === 'null' ||  value.toLowerCase() === 'nil'` with empty string `''`.

## Code changes
- CSVPatientExtractor now defines two unalterableColumns, familyName and givenName.
- BaseCSVExtractor now accepts `unalterableColumns` as a constructor argument and stores that value in local state.
- Adds tests of this new normalization functionality

# Testing guidance
- Ensure current tests pass.
- Try adding a NULL value to a CSV and see if it is appropriately replaced with an empty string (one option: replace gender with NULL/NIL in `patient-information.csv` and see if the dataAbsentReason appears in the output bundle)
- Ensure that givenName and familyName aren't scrubbed, even if their values are NULL/NIL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ A Node.js framework for extracting mCODE FHIR resources. All resources are profi
     - [Extraction Date Range](#extraction-date-range)
       - [CLI From-Date and To-Date (NOT recommended use)](#cli-from-date-and-to-date-not-recommended-use)
     - [Troubleshooting](#troubleshooting)
+      - [NULL/NIL values found and replaced with empty-strings](#nullnil-values-found-and-replaced-with-empty-strings)
       - [Byte Order Markers in CSV Files](#byte-order-markers-in-csv-files)
   - [Terminology and Architecture](#terminology-and-architecture)
     - [Glossary](#glossary)
@@ -164,6 +165,10 @@ npm start -- --entries-filter --from-date <YYYY-MM-DD> --to-date <YYYY-MM-DD> --
 ```
 
 ### Troubleshooting
+
+#### NULL/NIL values found and replaced with empty-strings
+
+When CSV files are provided containing NULL/NIL values, those values are treated as empty values and are translated into ''. Each Extractor, however, defines a set of `unalterableColumns` which will be immune from this NULL/NIL correction. All values that are corrected will produce a `debug`-level message, and can be seen by running the extractor with the debug flag set.
 
 #### Byte Order Markers in CSV Files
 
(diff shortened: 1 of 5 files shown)


jafeltra:
@Dtphelan1 just noticed this is going into `main`. Probably want this to go to `develop`.

julianxcarter (approved):
Looks good to me!

jafeltra:
Thanks for that change! Not to continue being a backseat reviewer (I just am noticing these things while making release notes), but did we want to make a note of what we are doing in the README? I know we talked about it, but maybe we decided against this since we evaluated which columns should be manipulated.

dmendelowitz:
I went in and changed a patients given name to "nil" to make sure it didn't get wiped but their name ended up being output as "null" which seems very weird. 1. the name isn't supposed to be changed and 2. how did it go from nil to null?

Dtphelan1:
2) looks like it might have to do with caseSensitivity issues – when we create our headers, we lowercase them all. So `familyName` and `familyname` won't match. Handling this now 😢

Dtphelan1:
@dmendelowitz should have taken care of 2. Going to look into 1) now...

Dtphelan1:
I don't seem to be able to replicate 1). Could you send me some screen grabs showing me what you were seeing exactly?

dmendelowitz:
nil_doe
result

dmendelowitz:
When the name was "Jane" in the csv it properly output "Jane Doe"

dmendelowitz:
It looks like your most recent commit fixed the "nil" -> "null" thing

jafeltra:
Ohhhh the `null Doe` thing might be my fault. I think when I updated the patient CSV to not require family or given name, I don't think I updated the template where it combines the two into that one `text` field. So I think the same thing would probably happen if you didn't include any value for `givenName` as well.

dmendelowitz:
@jafeltra Yeah, when givenName is blank, I get "null Doe" also, so I guess that's an issue with the patient template not @Dtphelan1's new code

dmendelowitz (approved):
I all my issues have been resolved

dmendelowitz:
I know I already approved but this thought just came to my mind. Since I just learned that the patient IDs csv is parsed without the CSVModule, is there any chance that that CSV will have a nil or null that needs to be interpreted as empty? (I hope not)

Dtphelan1:
1) Great question – I don't think we need to worry about this right now;
2) This does raise a bigger question about whether or not the client's should be aware of patient-mrns more generally. Not a problem for now, but something to think about more moving forward 😄

PR #121, opened 2021-05-19
process.exit moved to cli.js

julianxcarter:
# Summary
Delegates fatal error handling to the cli, rather than the `mcodeApp`.
## New behavior
When a fatal error is encountered in the `mcodeApp`, the running process is now closed from the cli rather than from within  the app. The cli also handles any logging that will result from that error.
## Code changes
1. Within `cli.js`, a new `runApp()` function is introduced. This function calls the `mcodeApp()` function from within a `try` block. If an error is encountered, this function will log the error and exit the node process. The `runApp()` function was necessary due to `mcodeApp()` being an asynchronous function, as calling an `async` function within a `try..catch` block can only be done from within an `async` function.
2. Within `app.js`, all fatal error handling within `mcodeApp()` has been removed. Errors are now thrown straight to the `runApp()` function where they are logged and the process is exited.
# Testing guidance
1. Ensure all current tests pass and that the extraction runs as expected with a simple `npm start`.
2. Manually test extraction with arguments that will trigger a fatal error, ensuring that the process still exits as expected and the error is appropriately logged. An example would be `npm start --  --entries-filter --from-date 54849/382/3742`, as this will trigger an error due to the invalid date.

### DIFF
--- a/src/cli/app.js
+++ b/src/cli/app.js
@@ -71,84 +71,78 @@ function getEffectiveFromDate(fromDate, runLogger) {
 }
 
 async function mcodeApp(Client, fromDate, toDate, pathToConfig, pathToRunLogs, debug, allEntries) {
-  try {
-    if (debug) logger.level = 'debug';
-    // Don't require a run-logs file if we are extracting all-entries. Only required when using --entries-filter.
-    if (!allEntries) checkLogFile(pathToRunLogs);
-    const config = getConfig(pathToConfig);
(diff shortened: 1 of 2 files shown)


Dtphelan1 (approved):
Stunning work, 10/10 would recommend. Great job!

dmendelowitz (approved):
Ditto what @Dtphelan1 said!

PR #126, opened 2021-06-03
Modified app.js to return object instead of writing to file; moved code to write to file to cli.js

SenecaAnderson:
# Summary
The function mcodeApp in src/cli/app.js was modified to return the extractedData object instead of writing the results to a file. src/cli/cli.js was modified to write the results of calling mocdeApp to a file.

## New behavior
cli.js writes the results of extraction to a file instead of app.js.

## Code changes
The code to write to a file was moved from app.js to cli.js and replaced with a return statement.

# Testing guidance
Use 'npm start' to run extraction.

### DIFF
--- a/src/cli/cli.js
+++ b/src/cli/cli.js
@@ -27,8 +27,7 @@ const allEntries = !entriesFilter;
 
 async function runApp() {
   try {
-    extractedData = await mcodeApp(MCODEClient, fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries);
-
+    const extractedData = await mcodeApp(MCODEClient, fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries);
 
     // Finally, save the data to disk
     const outputPath = './output';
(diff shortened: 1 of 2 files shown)


src/cli/cli.js
```diff
@@ -26,8 +27,24 @@ const allEntries = !entriesFilter;
 
 async function runApp() {
   try {
-    await mcodeApp(MCODEClient, fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries);
-  } catch (e) {
+    extractedData = await mcodeApp(MCODEClient, fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries);
```
mgramigna, line 30: Since this is a new variable, you'll need to add a `const extractedData = ` instead so Javascript knows

mgramigna (changes requested):

mgramigna:
I see there are some linter errors being caught by Github actions here. We have a pretty strict linter set up in the MEF, which allows us to enforce code style and make sure the syntax is correct.

You can configure VS Code to show you the lint errors by installing this extension: https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint

Once you do, you should see some errors in the editor that should need to be resolved. You can also check lint locally with the npm script we've added:

```
npm run lint
```

We also have a command that will fix and automatically fixable lint issues:

```
npm run lint-fix
```

mgramigna:
Other than the lint errors and the change request I made, this looks good and works as expected

mgramigna (approved):

jafeltra (approved):
Looks great!

PR #131, opened 2021-06-15
Add test for Patient Id CSV with BOM

jafeltra:
# Summary
Adds new test to ensure `parsePatientIds` properly handles CSVs with BOMs.

## New behavior
None

## Code changes
New test and fixture file

# Testing guidance
The new test should pass. Removing the `bom: true` option from the `parse` in `parsePatientIds` will cause the new test to fail, and adding it back will let it pass.


mgramigna (approved):
Confirmed that tests pass/look good. Confirmed that fixture CSV has BOMs. 10/10

Dtphelan1 (approved):

PR #133, opened 2021-06-18
Update csv config schema to include all properties

mgramigna:
# Summary

Fill out the rest of the JSON schema for CSV config

## New behavior

N/A

## Code changes

Add all properties and some required ones to JSON Schema

# Testing guidance

[here](https://www.jsonschemavalidator.net/s/NFT5L56B) is a URL with the updated JSON schema loaded into the validator.

Play around with this, paste your config in, delete some properties, invalid data type stuff, etc.

A visual inspection of the schema is also a good idea, it's not too dense

### DIFF
--- a/src/helpers/schemas/config.schema.json
+++ b/src/helpers/schemas/config.schema.json
@@ -7,6 +7,9 @@
     "patientIdCsvPath": {
       "type": "string"
     },
+    "commonExtractorArgs": {
+      "$ref": "#/$defs/commonExtractorArgs"
+    },
     "notificationInfo": {
       "$ref": "#/$defs/notificationInfo"
     },
@@ -18,6 +21,18 @@
     }
   },
   "$defs": {
+    "commonExtractorArgs": {
+      "type": "object",
+      "properties": {
+        "baseFhirUrl": {
+          "type": "string",
+          "format": "uri"
+        },
+        "requestHeaders": {
+          "type": "object"
+        }
+      }
+    },
     "notificationInfo": {
       "type": "object",
       "properties": {
@@ -32,8 +47,17 @@
           "type": "string"
         },
         "to": {
-          "type": "string",
-          "format": "email"
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            }
+          ]
         },
         "tlsRejectUnauthorized": {
           "type": "boolean"
@@ -75,8 +99,7 @@
         "type": {
           "type": "string"
         }
-      },
-      "required": ["filePath"]
+      }
     }
   },
   "required": [
--- a/src/helpers/schemas/config.schema.json
+++ b/src/helpers/schemas/config.schema.json
@@ -1,9 +1,15 @@
 {
   "$id": "csv-config",
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "description": "Schema for CSV config files. TODO: Fill out the rest of this schema will all config properties",
+  "description": "Schema for CSV config files",
   "type": "object",
   "properties": {
+    "patientIdCsvPath": {
+      "type": "string"
+    },
+    "notificationInfo": {
+      "$ref": "#/$defs/notificationInfo"
+    },
     "extractors": {
       "type": "array",
       "items": {
@@ -12,15 +18,69 @@
     }
   },
   "$defs": {
+    "notificationInfo": {
+      "type": "object",
+      "properties": {
+        "host": {
+          "type": "string",
+          "format": "hostname"
+        },
+        "port": {
+          "type": "integer"
+        },
+        "from": {
+          "type": "string"
+        },
+        "to": {
+          "type": "string",
+          "format": "email"
+        },
+        "tlsRejectUnauthorized": {
+          "type": "boolean"
+        }
+      },
+      "required": [
+        "host",
+        "port",
+        "from",
+        "to"
+      ]
+    },
     "extractor": {
       "type": "object",
       "properties": {
+        "label": {
+          "type": "string"
+        },
+        "type": {
+          "type": "string"
+        },
+        "constructorArgs": {
+          "$ref": "#/$defs/constructorArgs"
+        }
+      }
+    },
+    "constructorArgs": {
+      "type": "object",
+      "properties": {
+        "filePath": {
+          "type": "string"
+        },
+        "clinicalSiteID": {
+          "type": "string"
+        },
+        "clinicalSiteSystem": {
+          "type": "string"
+        },
         "type": {
           "type": "string"
         }
       },
-      "required": ["type"]
+      "required": ["filePath"]
     }
   },
-  "required": ["extractors"]
+  "required": [
+    "extractors",
+    "patientIdCsvPath"
+  ]
 }
--- a/src/helpers/schemas/config.schema.json
+++ b/src/helpers/schemas/config.schema.json
@@ -1,7 +1,7 @@
 {
   "$id": "csv-config",
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "description": "Schema for CSV config files",
+  "description": "Schema for mcode-extraction-framework config files",
   "type": "object",
   "properties": {
     "patientIdCsvPath": {
@@ -65,8 +65,6 @@
       },
       "required": [
         "host",
-        "port",
-        "from",
         "to"
       ]
     },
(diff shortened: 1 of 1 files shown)


src/helpers/schemas/config.schema.json
```diff
@@ -12,15 +18,69 @@
...
+        },
+        "clinicalSiteSystem": {
+          "type": "string"
+        },
         "type": {
           "type": "string"
         }
       },
-      "required": ["type"]
+      "required": ["filePath"]
```
Dtphelan1, line 79: DP1 - Given the work I'm doing, I don't think we necessarily want filePath to be required (when we have a URL, we won't need a filePath. 

Wondering aloud: is there some way of saying that there are three kinds of constructorArgs that you can see – for example, a fileSystem constructorArgs, a urlSystem constructorArgs, and an EHRSystem constructorArgs. For the initial work, I don't think we'll need something that complex. Just curious as it would allow us to enforce this requirement in the case where it is present, but ignore it in the case where it isn't. Writing it all out, I'm not even positive that a naïve parser would have a good way of distinguishing between the cases.... Just a #quirky Friday Five O'clock thought we can laugh at on Monday.

Dtphelan1:
DP2 - I think we could also use a commonExtractorArgs that defines an object with a baseFhirUrl property of type string with and format hostname

src/helpers/schemas/config.schema.json
```diff
@@ -12,15 +18,69 @@
...
+      "properties": {
+        "host": {
+          "type": "string",
+          "format": "hostname"
+        },
+        "port": {
+          "type": "integer"
+        },
+        "from": {
+          "type": "string"
```
Dtphelan1, line 32: DP3 - I think we want this to be format email too, maybe?

src/helpers/schemas/config.schema.json
mgramigna: thoughts?

src/helpers/schemas/config.schema.json
mgramigna: So I initially thought this too until I checked the example config and we include like a common name for the sender in addition to the email.

at least, this is what I have in my config:

image

mgramigna:
> DP2 - I think we could also use a commonExtractorArgs that defines an object with a baseFhirUrl property of type string with and format hostname

Clarifying Q: we make use of `commonExtractorArgs` with the URL properties (including auth and such) in other "proprietary" versions of the mcode-extraction-framework. Does that property also apply to CSV config? Since the schema I made is pretty specific to that.

src/helpers/schemas/config.schema.json
mgramigna: After talking with @jafeltra, it looks like the `to` portion can also be a comma separated list of emails rather than just one singular email. So maybe we should just use string for both for now and then later we can decide if we want any kind of smart regex checking?

Also, apparently `to` can also be an array, so this might be more complicated in the schema than we initially thought. I might replace it with this if that's ok:

```
        "to": {
          "anyOf": [
            {
              "type": "string"
            },
            {
              "type": "array",
              "items": {
                "type": "string"
              }
            }
          ]
        },
```

mgramigna:
> > DP2 - I think we could also use a commonExtractorArgs that defines an object with a baseFhirUrl property of type string with and format hostname
> 
> Clarifying Q: we make use of `commonExtractorArgs` with the URL properties (including auth and such) in other "proprietary" versions of the mcode-extraction-framework. Does that property also apply to CSV config? Since the schema I made is pretty specific to that.

Kind of answered my own question. Looks like we do parse/use commonExtractorArgs in the base apps so I will include a simple version of that in the schema

src/helpers/schemas/config.schema.json
Dtphelan1: I love the AnyOf usage here – it's a shame that we can't legally use the email format but if we have something like comma-separated strings then this definitely makes sense 😄

jafeltra: Darn I was going to suggest we could at least use the email format in the array, but Nodemailer supports an array of comma separated strings too. So we would have to choose to explicitly narrow what we accept compared to what Nodemailer accepts, which I'm not sure we want to do.

src/helpers/schemas/config.schema.json
```diff
@@ -1,9 +1,18 @@
 {
   "$id": "csv-config",
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "description": "Schema for CSV config files. TODO: Fill out the rest of this schema will all config properties",
+  "description": "Schema for CSV config files",
```
jafeltra, line 4: Since this isn't strictly used for only CSV extractors, should this be more generic for the mCODEApp? Or MEF?

src/helpers/schemas/config.schema.json
```diff
@@ -12,15 +21,89 @@
...
+        },
+        "tlsRejectUnauthorized": {
+          "type": "boolean"
+        }
+      },
+      "required": [
+        "host",
+        "port",
+        "from",
+        "to"
```
jafeltra, line 70: Technically `port` and `from` are not required. We default `from` to `mcode-extraction-errors@mitre.org` and nodemailer will default the port if it isn't provided. I'm not sure if we want to loosen the requirements here or if we would prefer people include these fields.

src/helpers/schemas/config.schema.json
```diff
@@ -12,15 +21,89 @@
...
+        },
+        "clinicalSiteID": {
+          "type": "string"
+        },
+        "clinicalSiteSystem": {
+          "type": "string"
+        },
+        "type": {
+          "type": "string"
+        }
```
jafeltra, line 101: I think that `mask` should be another property on `constructorArgs`, which is an array of strings.

src/helpers/schemas/config.schema.json
```diff
@@ -12,15 +21,89 @@
...
+      "properties": {
+        "filePath": {
+          "type": "string"
+        },
+        "clinicalSiteID": {
+          "type": "string"
+        },
+        "clinicalSiteSystem": {
+          "type": "string"
+        },
```
jafeltra, line 98: These two properties, plus `mask`, made me wonder if we talked about possible ways to specify which `type`s of extractors require which properties. `mask` only applies to the two patient extractors and `clinicalSiteID` and `clinicalSiteSystem` only apply to the clinical trial extractor. Is it too complicated/impossible to specify that level of requirement? If anything, it would make the schema more complicated than it is now, so maybe it isn't worth it, or maybe we already talked about that and decided it wasn't worth it and I just forgot.

src/helpers/schemas/config.schema.json
```diff
@@ -12,15 +21,89 @@
...
+      ]
+    },
     "extractor": {
       "type": "object",
       "properties": {
+        "label": {
+          "type": "string"
+        },
         "type": {
           "type": "string"
```
jafeltra, line 80: This is another thought that I can't remember if we have talked about previously, but is it possible to have an enumerated list of what `type`s we accept here? The types provided here should be limited to the extractors registered to the client. Or maybe that requires too much duplication and would mean every new extractor needs to be listed in the schema, which might be annoying, and we might rather keep the schema generic. I don't have a strong preference, but figured I'd ask since I got this far typing this lol

jafeltra (commented):
This looks good! I had a couple minor comments and a couple general questions.

src/helpers/schemas/config.schema.json
jafeltra: Sorry I missed where this was already discussed and decided against.

src/helpers/schemas/config.schema.json
mgramigna: Per discussion, separate task was made

src/helpers/schemas/config.schema.json
mgramigna: This thought might fall under Jules' recent comment which is now a Jira: to investigate tying "different formats" of schema to "different" kinds of things. E.g. constructorArgs may look one way for extractors of type X but different for extractors of type Y

jafeltra (approved):
This looks good to me! All my questions were addressed or made into jira tasks. I'm approving since I'll be out tomorrow, but happy to be a part of any conversations about the comments from Dylan on Thursday if you wait until then to discuss!

src/helpers/schemas/config.schema.json
Dtphelan1: I think that's fair. Definitely out of the scope of this I'd say. I think the AnyOf operator would be good here also, but not necessary right now in my opinion.

Dtphelan1 (approved):
After responding to my comments and addressing Jules' comments, this looks good to me. Great work Matt 😄

PR #135, opened 2021-06-23
Adding config file validator

dmendelowitz:
# Summary
The config  file is now checked for errors by validating it against the new config schema
## New behavior
When there are one or more errors in the config file, the client will fail and tell the user what is wrong with their config file
## Code changes
- `configValidator.js` contains the new `validateConfig()` function that performs validation
- `checkInputAndConfig()` in `app.js` now calls the new `validateConfig()` function instead of manually checking for `patientIdsCsvPath`
- Adjusted tests in `app.test.js` to work with the new validator
- Added new tests for the config validator in `configValidator.test.js`
# Testing guidance
- Ensure that when the config has no errors, the client still runs as expected
- Mess up the config file in more than one way (remove a required field, use a wrong type, use a wrong format) and make sure the error message alerts you of all errors in your config
- Make sure all tests still pass and new tests make sense

### DIFF
--- a/src/application/app.js
+++ b/src/application/app.js
@@ -7,6 +7,7 @@ const { sendEmailNotification, zipErrors } = require('./tools/emailNotifications
 const { extractDataForPatients } = require('./tools/mcodeExtraction');
 const { maskMRN } = require('../helpers/patientUtils');
 const { parsePatientIds } = require('../helpers/appUtils');
+const { validateConfig } = require('../helpers/configValidator');
 
 function getConfig(pathToConfig) {
   // Checks pathToConfig points to valid JSON file
@@ -20,7 +21,7 @@ function getConfig(pathToConfig) {
 
 function checkInputAndConfig(config, fromDate, toDate) {
   // Check input args and needed config variables based on client being used
-  const { patientIdCsvPath } = config;
+  validateConfig(config);
 
   // Check if `fromDate` is a valid date
   if (fromDate && !moment(fromDate).isValid()) {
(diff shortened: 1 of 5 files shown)


src/helpers/configValidator.js
```diff
@@ -0,0 +1,17 @@
...
+const metaSchema = require('ajv/lib/refs/json-schema-draft-06.json');
+const configSchema = require('./schemas/config.schema.json');
+
+const ajv = new Ajv({ logger: false, allErrors: true });
+ajv.addMetaSchema(metaSchema);
+const validator = ajv.addSchema(configSchema, 'config');
+
+function validateConfig(config) {
+  const valid = validator.validate('config', config);
+  const errors = ajv.errorsText(validator.errors, { dataVar: 'config' });
```
julianxcarter, line 11: I did not realize there was a way to access errors from the base `Ajv` object, good find!

julianxcarter (approved):
This works as intended and looks great to me

src/application/app.js
```diff
@@ -20,7 +21,7 @@ function getConfig(pathToConfig) {
 
 function checkInputAndConfig(config, fromDate, toDate) {
   // Check input args and needed config variables based on client being used
-  const { patientIdCsvPath } = config;
+  validateConfig(config);
```
mgramigna, line 24: I think it might be useful to have some (likely debug level?) log statements here to indicate that this is happening/has succeeded after the function call

mgramigna:
Made a tiny comment but everything about the functionality/tests looks good to me

mgramigna (approved):

PR #137, opened 2021-06-29
Moving MRN masking to maskPatientData

dmendelowitz:
# Summary
Now that other extractors get the Patient ID from context, MRN can just be masked during extraction instead of at the end of extraction and the masked ID will propagate to all other resources.
## New behavior
Behavior should be the same as before
## Code changes
- Post extraction MRN masking removed from `app.js`
- `maskMRN()` removed from `patientUtils.js`
- `maskPatientData()` updated to replace `id` and `fullUrl` with hashed version of MRN
- Relevant tests in `patientUtils.test.js` updated to reflect changes
# Testing guidance
Add `"mask": ["mrn"]` to `constructorArgs` of the Patient extractor and see that all instances of the MRN used as an ID are replaced with the hashed version of the MRN (appears in Patient's `id` and `fullUrl` fields and `subject` or `individual` references in other resources)


Dtphelan1 (approved):
This looks great! Nice work Dylan 😄

julianxcarter (approved):
Looks good to me!

PR #138, opened 2021-06-30
Adding custom schema format for comma separated emails

dmendelowitz:
# Summary
Adding a custom format to the config validator that checks if a string has the format of emails separated by commas, for example: `whoever@whatver.com,somone@wherever.com`. This new format is used to validate the `notificationInfo.to` field in the config
## New behavior
The config validator will now ensure that the `notificationInfo.to` field of the config is either an array of emails or a string of emails separated by commas
## Code changes
- Custom format, `comma-separated-emails`, added to `configValidator.js` which splits a string at any commas, strips excess whitespace and then validates the resulting strings against Ajv's email format regex
- `config.schema.json` updated to use `comma-separated-emails` for `notificationInfo.to`
# Testing guidance
Here are some variations to test:
- `test@example.com` a single email should still be valid
- `test#example.com` a single invalid email should not be valid
- `test@example.com,example@test.com` emails separated by comma should be valid
- `test@example.com,demo&example.com` should not be valid if any item is not a valid email
- Be sure to test any other cases you can think of in case I missed something

### DIFF
--- a/src/helpers/configValidator.js
+++ b/src/helpers/configValidator.js
@@ -5,6 +5,15 @@ const configSchema = require('./schemas/config.schema.json');
 
 const ajv = new Ajv({ logger: false, allErrors: true });
 ajv.addMetaSchema(metaSchema);
+
+ajv.addFormat('comma-separated-emails', {
+  type: 'string',
+  validate: (emails) => {
+    const emailRegex = new RegExp(/^[a-z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?(?:\.[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)*$/i);
+    return emails.split(',').every((email) => emailRegex.test(email.trim()));
+  },
+});
+
 const validator = ajv.addSchema(configSchema, 'config');
 
 function validateConfig(config) {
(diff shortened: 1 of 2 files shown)


dmendelowitz:
Also, I sort of just assumed it would be a good thing to strip excess whitespace, but does anyone actually know if notificationInfo.to needs to be strictly commas with no spaces?

src/helpers/configValidator.js
```diff
@@ -5,6 +5,15 @@ const configSchema = require('./schemas/config.schema.json');
 
 const ajv = new Ajv({ logger: false, allErrors: true });
 ajv.addMetaSchema(metaSchema);
+
+ajv.addFormat('comma-separated-emails', {
+  type: 'string',
+  validate: (emails) => {
+    const emailRegex = new RegExp(/^[a-z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?(?:\.[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)*$/i);
```
Dtphelan1, line 12: DP1 - Could we get a link to either the Stack Overflow or the AJV source of this RegExp

Dtphelan1 (approved):
This looks good. Tested with both an array of strings and a single string with multiple, commas-separated emails and things look good. Great work!

julianxcarter (approved):
Validation works flawlessly for me, both within the array and also in a comma-separated string. Nice work!

PR #139, opened 2021-07-01
Adding email with name custom schema format

dmendelowitz:
# Summary
`notificationInfo.from` now uses a custom format for emails with a name in front of them. This format considers anything that ends with a valid email to be valid, for example: `Example Name example@demo.com`
## New behavior
Config validator has a new custom format that splits the input on spaces and checks that the last element is a valid email.
## Code changes
- `configValidator.js` now has new custom format `email-with-name` described above
- `config.schema.json` updated to use new format for `notificationInfo.from`
# Testing guidance
Cases to test:
- `Any number of words followed by anEmail@test.com` should be valid
- `justAnEmail@test.com` should be valid
- `Words with no email` should not be valid
- `emailFirst@test.com then some words` should not be valid

### DIFF
--- a/src/helpers/configValidator.js
+++ b/src/helpers/configValidator.js
@@ -19,7 +19,7 @@ ajv.addFormat('email-with-name', {
   validate: (email) => {
     // this is Ajv's regex for email format (https://github.com/ajv-validator/ajv-formats/blob/master/src/formats.ts#L106)
     const emailRegex = new RegExp(/^[a-z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?(?:\.[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)*$/i);
-    return emailRegex.test(email.split(' ').pop());
+    return emailRegex.test(email.trim().split(' ').pop());
   },
 });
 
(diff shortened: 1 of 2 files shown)


src/helpers/configValidator.js
```diff
@@ -14,6 +14,14 @@ ajv.addFormat('comma-separated-emails', {
     return emails.split(',').every((email) => emailRegex.test(email.trim()));
   },
 });
+ajv.addFormat('email-with-name', {
+  type: 'string',
+  validate: (email) => {
+    // this is Ajv's regex for email format (https://github.com/ajv-validator/ajv-formats/blob/master/src/formats.ts#L106)
+    const emailRegex = new RegExp(/^[a-z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?(?:\.[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)*$/i);
+    return emailRegex.test(email.split(' ').pop());
```
Dtphelan1, line 22: DP1 - We might want to trim this email before splitting. If there's empty space at the end of the string I think that will cause an error right? Console test here: 
image

Dtphelan1:
Beyond my one small nitpick, this looks great!

Dtphelan1 (approved):
Great work Dylan!

julianxcarter (approved):
Looks great!

PR #141, opened 2021-07-15
Better logging in the CSVURLModule 

Dtphelan1:
# Summary

This PR addresses STEAM-670: 'CSVURLModule should validate the the URL it receives is a valid URL'. Upon closer inspection, it doesn't look like we do the level of validation we expected in the CSVFileModule. Accordingly, all this task takes up is better logging in fillDataCache logic. 

## New behavior

When a URL is provided that cannot be used successfully in CSV extraction, error messages are logged to the console. Additionally, debug messages have been added throughout the `fillDataCache` routine. 

## Code changes
- Logs added throughout the `fillDataCache` function.

# Testing guidance
- With a URL-enabled csv extractor, run the tool with a valid URL and ensure that extraction works. 
- With a URL-enabled csv extractor, run the tool with an invalid URL and ensure that the error message logs as expected. 

Below is an example (courtesy of @julianxcarter ) of a valid URL extractor config
```json
{
      "label": "cancerDiseaseStatus",
      "type": "CSVCancerDiseaseStatusExtractor",
      "constructorArgs": {
        "url": "https://raw.githubusercontent.com/mcode/mcode-extraction-framework/main/test/sample-client-data/cancer-disease-status-information.csv"
      }
    }
```

### DIFF
--- a/src/client/BaseClient.js
+++ b/src/client/BaseClient.js
@@ -43,19 +43,23 @@ class BaseClient {
     // Using Reduce to compute the validity of all extractors
     const allExtractorsValid = await this.extractors.reduce(async (curExtractorsValid, curExtractor) => {
       const { name } = curExtractor.constructor;
-
       if (curExtractor.validate) {
         logger.debug(`Validating ${name}`);
-        const isExtractorValid = await curExtractor.validate();
-        if (isExtractorValid) {
-          logger.debug(`Extractor ${name} PASSED CSV validation`);
-        } else {
+        try {
+          const isExtractorValid = await curExtractor.validate();
+          if (isExtractorValid) {
+            logger.debug(`Extractor ${name} PASSED CSV validation`);
+          } else {
+            logger.warn(`Extractor ${name} FAILED CSV validation`);
+          }
+          return ((await curExtractorsValid) && isExtractorValid);
+        } catch (e) {
           logger.warn(`Extractor ${name} FAILED CSV validation`);
+          return false;
         }
-        return (curExtractorsValid && isExtractorValid);
       }
       return curExtractorsValid;
-    }, true);
+    }, Promise.resolve(true));
 
     if (allExtractorsValid) {
       logger.info('Validation succeeded');
--- a/src/client/BaseClient.js
+++ b/src/client/BaseClient.js
@@ -54,7 +54,7 @@ class BaseClient {
           }
           return ((await curExtractorsValid) && isExtractorValid);
         } catch (e) {
-          logger.warn(`Extractor ${name} FAILED CSV validation`);
+          logger.warn(`Extractor ${name} could not validate. Encountered the following error: ${e.message}`);
           return false;
         }
       }
--- a/src/modules/CSVURLModule.js
+++ b/src/modules/CSVURLModule.js
@@ -20,7 +20,7 @@ class CSVURLModule {
       const csvData = await axios.get(this.url)
         .then((res) => res.data)
         .catch((e) => {
-          logger.error('Error occurred when getting CSV data using url');
+          logger.error('Error occurred when making a connection to this url');
           throw e;
         });
       logger.debug('Web request successful');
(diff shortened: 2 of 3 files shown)


src/modules/CSVURLModule.js
```diff
@@ -16,13 +16,22 @@ class CSVURLModule {
...
+        });
+      logger.debug('Web request successful');
       // Parse then normalize the data
       const parsedData = parse(csvData, {
         columns: (header) => header.map((column) => stringNormalizer(column)),
         bom: true,
       });
+      logger.debug('Data parsing successful');
       this.data = normalizeEmptyValues(parsedData, this.unalterableColumns);
+      logger.debug('Normalization of empty values successful');
```
dmendelowitz, line 34: Should this message just be inside the `normalizeEmptyValues` function so it shows up for both File and URL modules? Alternatively, is this redundant since the `normalizeEmptyValues` function will already log a warning when it does it's thing?

dmendelowitz (commented):
This looks good to me! Just had one minor comment

src/modules/CSVURLModule.js
Dtphelan1: That's a totally fair point. I've addressed this in my most recent commit

dmendelowitz (approved):

src/modules/CSVURLModule.js
```diff
@@ -16,12 +16,20 @@ class CSVURLModule {
   // If data is already cached, this function does nothing
   async fillDataCache() {
     if (!this.data) {
-      const csvData = await axios.get(this.url).then((res) => res.data);
+      logger.debug('Filling the data cache of CSVURLModule');
+      const csvData = await axios.get(this.url)
+        .then((res) => res.data)
+        .catch((e) => {
+          logger.error('Error occurred when getting CSV data using url');
+          throw e;
```
jafeltra, line 24: When I run this with an invalid URL, I get a `UnhandledPromiseRejectionWarning` warning logged (in addition to the appropriate log messages). If I comment out this line, I end up with an error from the `csvValidator` and the same `UnhandledPromiseRejectionWarning` is logged through that, and if I comment out the full catch block, the warning comes from the thrown error that isn't being caught. I wanted to see if you had also seen this warning and what you thought about it., but since it seems like I can run into this in a case we previously had, maybe this has already been discussed.

jafeltra (commented):
The changes here make sense and I get the appropriate log messages. I have one broader question that I'm not sure if was discussed previously or not.

src/modules/CSVURLModule.js
Dtphelan1: Jules and I discussed this offline. While the root cause of this issue doesn't have to do with my changes per se, I'm going to look into it briefly 😄

Dtphelan1: The most recent commit should have addressed the problem you were seeing beefore. Let me know if this is still an issue or if there is anything else you want to discuss.

src/client/BaseClient.js
```diff
@@ -43,19 +43,23 @@ class BaseClient {
...
+        try {
+          const isExtractorValid = await curExtractor.validate();
+          if (isExtractorValid) {
+            logger.debug(`Extractor ${name} PASSED CSV validation`);
+          } else {
+            logger.warn(`Extractor ${name} FAILED CSV validation`);
+          }
+          return ((await curExtractorsValid) && isExtractorValid);
+        } catch (e) {
           logger.warn(`Extractor ${name} FAILED CSV validation`);
```
jafeltra, line 57: Two thoughts in one suggestion:

- With the log message here, we don't get any information about the error that gets thrown in the CSVURLModule. I think it would be nice to include that information. The first line in the suggestion is one way to address this.
- But then when you just do the first line, the generic error gets thrown on line 67 and it feels not ideal to lose the stack trace info of the real error. The second line to throw the error from the catch admittedly feels odd but also lets the final debug message have the stack track to the 404 error.

```suggestion
          logger.warn(`Extractor ${name} could not validate. Encountered the following error: ${e.message}`);
          throw e;
```

jafeltra (approved):
Thanks for making those changes!

jafeltra:
@dmendelowitz do you still approve with the minor changes made since you last approved? Just wanted to give you a chance to take a look before I merged!

PR #142, opened 2021-07-15
Moving config file parsing from app to CLI

dmendelowitz:
# Summary
Rather than pass the config file path to `mcodeApp`, the config will now be parsed in the CLI and passed to the app as an object.
## New behavior
The client should behave as before
## Code changes
- Renamed `configValidator.js` to `configUtils.js` and moved `getConfig()` from `app.js` to there
- Changed the `pathToConfig` argument of `mcodeApp` to `config` and removed config parsing from that file
- Added `getConfig()` call to `cli.js`, the result is then passed into `mcodeApp`
- Tests for `getConfig()` moved from `app.test.js` to the newly renamed 'configUtils.test.js'
# Testing guidance
- Make sure extraction still works as expected
- Make sure config parsing still works as expected (errors still thrown for bad paths to config, etc)


Dtphelan1:
This looks really good!

The only thing I can think of is exporting the `getConfig` and `validateConfig` functions in the main export file. I imagine that those will be useful in the other repos.

Dtphelan1 (approved):

mgramigna (approved):
Looks great

PR #144, opened 2021-08-02
Expand VS for Conditions

Dtphelan1:
# Summary

This PR addresses STEAM-423. CT Matching task is using snomed condition codes. MEF looks to see if condition codes are in primary or secondary value set in order to properly add category.coding values to conform to mCODE cancer condition profiles. However, because our VS files are not expanded, we don't properly satisfy the 'isCancerCondition' check, so we don't add the necessary category codings. 

This PR expands all mCODE VS's that have `include` objects which define codes intensionally with a `filter` object. 

## New behavior

No new behavior, just new codes against which we match for `isCancerCondition` function calls

## Code changes

None.

# Testing guidance

Before checking out this PR, run the extraction tool and inspect the Condition resource and confirm that the condition's do not have a category code with value "64572001" (this category code gets added when the condition is recognized as a Cancer condition). 

Afterwards, checkout the branch and run extraction again. Confirm that this category code, "64572001", is now showing up on condition resources. 

*NOTE*: Before merging, we should remove the old VS's in the `valuesets/original` dir. I've left them in so people can see the change that's taken place in expansion.


dmendelowitz (approved):
This looks good to me!

PR #145, opened 2021-08-05
Replaced shajs with crypto

Dtphelan1:
# Summary

Removes the shajs dependency and uses the Node-native Crypto library. 

## New behavior

Ideally none.

## Code changes

Replaces all instances of the sha library with the crypto library. 

# Testing guidance

Ensure all tests pass.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9539,15 +9539,6 @@
         }
       }
     },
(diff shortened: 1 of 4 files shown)


dmendelowitz (approved):
Seamless!

PR #148, opened 2021-08-10
Updating CancerDiseaseStatusTemplate to STU2

dmendelowitz:
# Summary
CancerDiseaseStatus profile now uses a newly-created LOINC code, 97509-4 “Cancer Disease Progression”. In STU 1, the LOINC code chosen to represent this observation (88040-1, “Response to cancer treatment”) did not precisely match the meaning of this profile because cancer disease status is observable regardless of whether the patient is under treatment
## New behavior
Cancer Disease Status template now uses the new LOINC code from STU2
## Code changes
- `CancerDiseaseStatusTemplate.js` now uses the new LOINC code
- Relevant test fixtures updated to match the new code
# Testing guidance
- Ensure that all tests pass
- Run the client with CancerDiseaseStatus and see that the output now contains the new LOINC code: `97509-4`
- Check if there are any updates from STU2 for the CancerDiseaseStatusTemplate that I missed

### DIFF
--- a/src/templates/CancerDiseaseStatusTemplate.js
+++ b/src/templates/CancerDiseaseStatusTemplate.js
@@ -74,8 +74,8 @@ function cancerDiseaseStatusTemplate({
       coding: [
         coding({
           system: 'http://loinc.org',
-          code: '88040-1',
-          display: 'Response to cancer treatment',
+          code: '97509-4',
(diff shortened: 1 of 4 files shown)


jafeltra (approved):
This looks great! I didn't notice any other changes in the profile!

PR #151, opened 2021-08-20
Updating condition template to STU2

dmendelowitz:
# Summary
Category code is no longer fixed for Primary and Secondary Cancers. New profiles https://build.fhir.org/ig/HL7/fhir-mCODE-ig/branches/master/StructureDefinition-mcode-primary-cancer-condition.html
## New behavior
Previously, cancer conditions had 2 category codings, one being a fixed snomed code. In STU2 the fixed code is no longer there and now there is just the US core condition code
## Code changes
- Removed the fixed snomed code from `categoryArrayTemplate()` in `ConditionTemplate.js`
- Updated relevant tests and test fixtures
# Testing guidance
- Make sure all tests still pass
- When running extraction with the condition extractor, cancer conditions should now only have 1 coding in the category array, a US core condition code


Dtphelan1:
One super small DP1 – on https://github.com/mcode/mcode-extraction-framework/blob/1fe0b4acbdc301f36ead79e66caca9e6c4c45d05/src/templates/ConditionTemplate.js#L116, we should probably remove the second, now-unnecessary argument to the `categoryArray` fn.

dmendelowitz:
I'll make the fix for DP1 soon, but another thing I noticed is that the `isConditionCodeCancer()` function in `conditionUtils.js` is now not used anywhere in the MEF. Should I remove that? Is that used in any other repos? Maybe it's just good to keep around for possible future use?

Dtphelan1:
DP2 - There are a few places in the `condition.test.js` file that also need a bit of updating. I'll outline them in individual suggestions, but group them all under this tag since they're related to the removal of the cancer-condition test case. 

After that I think I'm ready to approve. This overall looks great to me 😄

Dtphelan1:
> I'll make the fix for DP1 soon, but another thing I noticed is that the `isConditionCodeCancer()` function in `conditionUtils.js` is now not used anywhere in the MEF. Should I remove that? Is that used in any other repos? Maybe it's just good to keep around for possible future use?

I think for now let's keep that fn around, since I anticipate we may need it at a later date. Not the best practice to keep stale code around, but the idea of knowing if something is or is not a cancer is valuable enough in mCODE-land to warrant keeping it around 😄

test/templates/condition.test.js
```diff
@@ -115,22 +114,6 @@ describe('test Condition template', () => {
...
-      ],
-    };
-    // Use a cancer condition code when generating resource
-    const cancerConditionData = { ...CONDITION_MINIMAL_DATA, code: { code: 'C02.0', system: 'http://hl7.org/fhir/sid/icd-10-cm' } };
-    const generatedCondition = conditionTemplate(cancerConditionData);
-    expect(generatedCondition.category).toContainEqual(diseaseCategory);
-    expect(generatedCondition).toEqual(cancerValidExampleCondition);
-  });
-
   test('non-cancer conditions should not include the "disease" category', () => {
```
Dtphelan1, line 117: DP2 - I think we can get rid of this entire test case. We don't really need to check category-length anymore, and we already check that rendering a conditionTemplate produces an expected condition on line 76. 

I thought there would be more, but this should be it for DP2.

Dtphelan1 (approved):

Dtphelan1:
Looks good to me!

PR #153, opened 2021-09-13
CancerRelatedMedication STU2 Updates

dmendelowitz:
# Summary
CancerRelatedMedicationExtractor has been converted to CancerRelatedMedicationAdministrationExtractor and a new CancerRelatedMedicationRequestExtractor has been added. 

Profiles:
https://build.fhir.org/ig/HL7/fhir-mCODE-ig/branches/master/StructureDefinition-mcode-cancer-related-medication-request.html
http://hl7.org/fhir/us/mcode/2021May/StructureDefinition-mcode-cancer-related-medication-administration.html

## New behavior
CancerRelatedMedicationAdministrationExtractor behaves the same as the old CancerRelatedMedicationExtractor. It's pretty much just a renaming. CancerRelatedMedicationRequestExtractor is a new extractor based on the profile from STU2 linked above.
## Code changes
- All files associated with the previous `CancerRelatedMedicationExtractor` (the extractor, template and tests) have been renamed to  `CancerRelatedMedicationAdministration` to distinguish them from `CancerRelatedMedicationRequest`
- New extractor, template and tests have been added for `CancerRelatedMedicationRequest` profile
- `MCODEClient.js`, `index.js` and `ResourceGenerator.js` updated to support the new and renamed extractors
- Updated sample client data to have new `cancer-related-medication-request-information.csv` and renamed `cancer-related-medication-administration-information.csv`

# Testing guidance
- Ensure that in converting CancerRelatedMedicationStatement to CancerRelatedMedicationAdministration, I didn't miss any other changes made in STU2
- Check that my implementation of CancerRelatedMedicationRequest properly implements the profile (if I am missing any fields or implementing any fields that aren't necessary) and check that the new CSV format makes sense
- Ensure that the client runs and tests pass
- Here is an example config for the relevant extractors:
```
"extractors": [
    {
      "label": "patient",
      "type": "CSVPatientExtractor",
      "constructorArgs": {
        "filePath": "./test/sample-client-data/patient-information.csv"
      }
    },
    {
      "label": "cancerRelatedMedicationAdministration",
      "type": "CSVCancerRelatedMedicationAdministrationExtractor",
      "constructorArgs": {
        "filePath": "./test/sample-client-data/cancer-related-medication-administration-information.csv"
      }
    },
    {
      "label": "cancerRelatedMedicationRequest",
      "type": "CSVCancerRelatedMedicationRequestExtractor",
      "constructorArgs": {
        "filePath": "./test/sample-client-data/cancer-related-medication-request-information.csv"
      }
    }
  ]
```


jafeltra (commented):
This looks really great! All the changes in the code and the new fields in the extractor look correct to me! I ran the FHIR validator against the output and found a few things we might want to update in our sample data, so I left a few comments related to that, along with just a couple quick general comments.

JA1 - The `src/index.js` file should be updated similarly to the `src/extractors/index.js` file to include the new extractors.

JA2 - Can you update the `csv.config.example.json` file to use the two new extractors instead of the one old one?

JA3 - It looks like the status values for MedicationAdministration are different than they were for MedicationStatement. The options it can be are in [this valueset](http://hl7.org/fhir/R4/valueset-medication-admin-status.html). Do you think you could update the values we have in the sample-client-data CSV just so they reflect the expected values?

JA4 - Similar to above, the values for the reasonCode in the MedicationAdministration don't seem to match up with codes in the current [Cancer Disorder Value Set](https://build.fhir.org/ig/HL7/fhir-mCODE-ig/ValueSet-mcode-cancer-disorder-vs.html). Could you update the values in our sample data fields to match codes in that value set?

JA5 - Pretty much the same as JA4, but the reasonCode for the MedicationRequest in the sample CSV also doesn't match a code in the current value set. Can you update that as well?

Dtphelan1:
This looks great Dylan! In addition to the comments Julia made: 
DP1 - In the same way that example data was added to the `test/sample-client-data`, we should update similar files in the `docs` folder. Specifically, let's get rid of the `docs/cancer-related-medication.csv` and add `docs.cancer-related-medication-request.csv`+ `docs.cancer-related-medication-administration.csv` documentation files.
DP2 - Similar updates, but in the CSV Templates file. 
DP3 - Minor note: cancerRelatedMedicationAdministrationExtractor.test.js and cancerRelatedMedicationRequestExtractor.test.js use variables called`csvCancerRelatedMedicationExtractor`, which might make life easier from a shorthand perspective writing-wise but might be nice to replace with the xAdmin or xRequest equivalent.

Dtphelan1:
DP4  - For some of the common template snippets – `treatmentIntentTemplate`, `medicationTemplate`, and `subjectTemplate` come to mind – could we put those in separate Snippet files, then import them into the two main templates we're defining? Willing to push this into a separate task if it's sufficiently hairy/out of scope.

jafeltra (commented):
This looks good! All the JAs are resolved! The one last thing that I noticed is that I don't think you need to add the `~$CSV_Templates.xlsx` file. I think the regular file is sufficient.

jafeltra (approved):
Looks great!

Dtphelan1 (approved):
This looks good to me Dylan – addresses all the comments I made eariler. Great work Dylan!

PR #155, opened 2021-09-29
New options for Patient masking

julianxcarter:
# Summary
Makes the following elements in the Patient resource maskable: `telecom`, `multipleBirth` (both `multipleBirthBoolean` and `multipleBirthInteger`), `photo`, `contact`, `generalPractitioner`, `managingOrganization`, and  `link`.
## New behavior
No changed behavior. The above elements are now supported as options within the `mask` array of the PatientExtractor config.
## Code changes
1. Support for new fields added to the `maskPatientData` function.
2. A new text fixture (`extended-patient-bundle.json`) was added to test masking, given that the newly supported fields are not supported by the CSVExtractors and will only ever occur in resources obtained via a FHIR search.
3. Updated readme to include the new masking fields.
# Testing guidance
1. Ensure that updated unit tests pass.
2. Ensure that the updated masking test fixture has elements consistent with what would be expected on the [CancerPatient resource](https://build.fhir.org/ig/HL7/fhir-mCODE-ig/StructureDefinition-mcode-cancer-patient.html) (namely that `_`s are placed before primitive elements, and that array elements have their dataAbsentReason extension contained within an array).
3. Ensure that the updated mask fixture is valid FHIR. This can be done using the [Inferno FHIR Validator](https://inferno.healthit.gov/validator/)

### DIFF
--- a/src/helpers/patientUtils.js
+++ b/src/helpers/patientUtils.js
@@ -78,7 +78,9 @@ function getPatientName(name) {
  * dataAbsentReason extension with value 'masked'
  * @param {Object} bundle a FHIR bundle with a Patient resource
  * @param {Array} mask an array of fields to mask. Values can be:
- * ['gender','mrn','name','address','birthDate','language','ethnicity','birthsex','race']
+ * 'gender','mrn','name','address','birthDate','language','ethnicity','birthsex',
+ * 'race', 'telecom', 'multipleBirth', 'photo', 'contact', 'generalPractitioner',
+ * 'managingOrganization', and 'link'
  */
 function maskPatientData(bundle, mask) {
   // get Patient resource from bundle
@@ -87,7 +89,24 @@ function maskPatientData(bundle, mask) {
     'Bundle.entry.where(resource.resourceType=\'Patient\').resource,first()',
   )[0];
 
-  const validFields = ['gender', 'mrn', 'name', 'address', 'birthDate', 'language', 'ethnicity', 'birthsex', 'race'];
+  const validFields = [
+    'gender',
+    'mrn',
+    'name',
+    'address',
+    'birthDate',
+    'language',
+    'ethnicity',
+    'birthsex',
+    'race',
+    'telecom',
+    'multipleBirth',
+    'photo',
+    'contact',
+    'generalPractitioner',
+    'managingOrganization',
+    'link',
+  ];
   const masked = extensionArr(dataAbsentReasonExtension('masked'));
 
   mask.forEach((field) => {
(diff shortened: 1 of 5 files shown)


src/helpers/patientUtils.js
```diff
@@ -157,6 +176,32 @@ function maskPatientData(bundle, mask) {
...
+    } else if (field === 'multipleBirth') {
+      if ('multipleBirthBoolean' in patient) {
+        delete patient.multipleBirthBoolean;
+        patient._multipleBirthBoolean = masked;
+      } else if ('multipleBirthInteger' in patient) {
+        delete patient.multipleBirthInteger;
+        patient._multipleBirthInteger = masked;
+      }
+    } else if (field === 'photo' && 'photo' in patient) {
+      delete patient.phoo;
```
Dtphelan1, line 191:
```suggestion
      delete patient.photo;
```
Superrrr small typo

Dtphelan1:
Overall this looks great to me. The one other thing I'll remark on is the resulting bundle is failing validation according to [Inferno's online validator](https://inferno.healthit.gov/validator). The errors I see are screncapped below. That said, I feel confident that our modelling of data-absent-reason for primitive values is correct [based on FHIR's documentation](https://www.hl7.org/fhir/json.html#primitive). I raise the point in case I'm 1) getting different validator results than you, 2) missing another aspect of this problem that might be interesting. 

tl;dr once my super small typo above is addressed, I'm good to approve; modelling data-absent extensions on primitives is hard; validating against them is harder.

Dtphelan1:
Validation errors/warnings image

src/helpers/patientUtils.js
julianxcarter: Good catch! Addressed in commit `713e4ce`

julianxcarter:
> Overall this looks great to me. The one other thing I'll remark on is the resulting bundle is failing validation according to [Inferno's online validator](https://inferno.healthit.gov/validator). The errors I see are screncapped below. That said, I feel confident that our modelling of data-absent-reason for primitive values is correct [based on FHIR's documentation](https://www.hl7.org/fhir/json.html#primitive). I raise the point in case I'm 1) getting different validator results than you, 2) missing another aspect of this problem that might be interesting.
> 
> tl;dr once my super small typo above is addressed, I'm good to approve; modelling data-absent extensions on primitives is hard; validating against them is harder.

I got the same errors and warnings when I ran the validation myself. They seem to be unavoidable for us in the sense that the elements causing them do actually require the properties being detected by the validator, but that including those required bits does amount to PII that's being masked in the first place. Some of these errors could be avoided by including some nebulously appropriate values for some of these elements (perhaps included `"Masked"` as a value for `name.given` and `name.family`, `"unknown"` as a coded value for [`gender`](http://hl7.org/fhir/R4/valueset-administrative-gender.html), or `"UNK"` as a coded value for the [`birthSex` extension](http://hl7.org/fhir/us/core/STU3.1/ValueSet-birthsex.html)). This would inch us closer to valid FHIR in the technical sense, but it's not clear to me that these codes and values are meant to be used in that way. Unfortunately I think in regards to masking we're gonna have to make a choice between semantically correct FHIR, and FHIR that follows the specification to a T.

Dtphelan1:
> > Overall this looks great to me. The one other thing I'll remark on is the resulting bundle is failing validation according to [Inferno's online validator](https://inferno.healthit.gov/validator). The errors I see are screncapped below. That said, I feel confident that our modelling of data-absent-reason for primitive values is correct [based on FHIR's documentation](https://www.hl7.org/fhir/json.html#primitive). I raise the point in case I'm 1) getting different validator results than you, 2) missing another aspect of this problem that might be interesting.
> > tl;dr once my super small typo above is addressed, I'm good to approve; modelling data-absent extensions on primitives is hard; validating against them is harder.
> 
> I got the same errors and warnings when I ran the validation myself. They seem to be unavoidable for us in the sense that the elements causing them do actually require the properties being detected by the validator, but that including those required bits does amount to PII that's being masked in the first place. Some of these errors could be avoided by including some nebulously appropriate values for some of these elements (perhaps included `"Masked"` as a value for `name.given` and `name.family`, `"unknown"` as a coded value for [`gender`](http://hl7.org/fhir/R4/valueset-administrative-gender.html), or `"UNK"` as a coded value for the [`birthSex` extension](http://hl7.org/fhir/us/core/STU3.1/ValueSet-birthsex.html)). This would inch us closer to valid FHIR in the technical sense, but it's not clear to me that these codes and values are meant to be used in that way. Unfortunately I think in regards to masking we're gonna have to make a choice between semantically correct FHIR, and FHIR that follows the specification to a T.

To be clear, I think what we have so far is perfect and the ideal. Just bizarre that validators don't seem keen on handling the else-if case where, if that property cannot be found, then check to see if there is an appropriate data-absent-reason. Might be something to discuss with the inferno team at some point. 

As far as this goes, I'm good to approve

Dtphelan1 (approved):

PR #157, opened 2021-10-05
updated schema; added titles to each possible to format

Dtphelan1:
Minor updates to our config file schema. Updates the `.notificationInfo.to` property so that: 
1. The AnyOf is changed to OneOf, which is more in line with what we want. For detailed explanations on the difference between these, see [here](https://json-schema.org/understanding-json-schema/reference/combining.html).
2. Each of the OneOf options have a title associated with them. This improves the result of rendering these options by React JSON Schema Form, which is used by the extraction-ui. A companion PR exists there and also needs to be reviewed.

Testing here should ensure that everything works as it previously did. Additionally, before merging these changes, we should review the extraction-ui and ensure that the visual changes look as desired.

### DIFF
--- a/src/helpers/schemas/config.schema.json
+++ b/src/helpers/schemas/config.schema.json
@@ -60,12 +60,12 @@
           "title": "To",
           "oneOf": [
             {
-              "title": "To, Comma Separated Emails",
+              "title": "To (comma separated list of emails)",
               "type": "string",
               "format": "comma-separated-emails"
             },
             {
-              "title": "To, Array of Emails",
+              "title": "To (list of emails)",
               "type": "array",
               "items": {
                 "type": "string",
--- a/src/helpers/schemas/config.schema.json
+++ b/src/helpers/schemas/config.schema.json
@@ -58,12 +58,14 @@
         },
         "to": {
           "title": "To",
-          "anyOf": [
+          "oneOf": [
             {
+              "title": "To, Comma Separated Emails",
               "type": "string",
               "format": "comma-separated-emails"
             },
             {
+              "title": "To, Array of Emails",
               "type": "array",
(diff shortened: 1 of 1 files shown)


src/helpers/schemas/config.schema.json
```diff
@@ -58,12 +58,14 @@
...
           "title": "To",
-          "anyOf": [
+          "oneOf": [
             {
+              "title": "To, Comma Separated Emails",
               "type": "string",
               "format": "comma-separated-emails"
             },
             {
+              "title": "To, Array of Emails",
```
jafeltra, line 68: Since this is going to be seen by less technical users, I wonder if this should be "list" rather than "array".

src/helpers/schemas/config.schema.json
```diff
@@ -58,12 +58,14 @@
         },
         "to": {
           "title": "To",
-          "anyOf": [
+          "oneOf": [
             {
+              "title": "To, Comma Separated Emails",
```
jafeltra, line 63: Similarly, maybe this can say "comma separated list of emails" just to make it a little clearer to a non-technical user.

jafeltra: For both of the titles, when I look at it in the UI, I feel like I see the two options as two different "To" items to be filled other, rather than two versions of the same field. I'm not sure if this would help, but would separating out and lowercasing the type descriptions help with this? I was thinking something like

```
To (comma separated list of emails)
```

I'm not sure if this will make any difference or if I'm just overthinking the labels and they're fine how they are, so I'm curious about your opinion on this!

jafeltra (commented):
Very minor comments about the wording, but this works as expected in the UI app!

PR #158, opened 2021-10-07
"all" option for Patient masking now supported in the config

julianxcarter:
# Summary
Adds support for a string value of `all` within the Patient Extractor constructorArgs. 
## New behavior
A string with a value of `all` can now be included in the Patient Extractor config to mask all supported Patient fields.
## Code changes
1. The mask property in `config.schema.json` has been updated to allow both strings and arrays.
2. New format added to `configUtils.js` to validate that any string included in the `mask` field is the `all` flag.
2. Support for the `all` flag added to the `FHIRPatientExtractor` and `CSVPatientExtractor`.
# Testing guidance
1. Ensure that using the `all` flag in the config actually masks all properties.
2. Ensure that including any other string as a value for the `mask` property in the config causes config validation to fail.
3. Ensure that the current array-based option still works as expected.

### DIFF
--- a/src/extractors/CSVPatientExtractor.js
+++ b/src/extractors/CSVPatientExtractor.js
@@ -9,26 +9,6 @@ const { getEmptyBundle } = require('../helpers/fhirUtils');
 const { CSVPatientSchema } = require('../helpers/schemas/csv');
 const logger = require('../helpers/logger');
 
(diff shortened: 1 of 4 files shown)


jafeltra (approved):
This looks great! I really like the refactor and masterful git work!

PR #162, opened 2021-11-04
dateTimeStamp now localizes to UTC

Dtphelan1:
In addressing the root cause of broken E-MEF PerformanceStatus tests, we realized that the failing tests were due to the fact that our formatDateTime function was always interpretting timestamps according to the local timezone. This is according to the `moment()` constructor specification [found here](https://momentjs.com/docs/#/parsing/). Also found in the moment parsing documentation is guidance on the `moment.parseZone()` function, which will maintain the provided timezone in the parsing of a string and default to UTC if none is provided. 

This PR updates our use of moment in our formatDate functions to use this newly discovered method, leading to more robust testing fixtures on the E-MEF that don't depend on a testing machine running in the Eastern Time Zone

### DIFF
--- a/src/helpers/dateUtils.js
+++ b/src/helpers/dateUtils.js
@@ -3,10 +3,10 @@ const logger = require('./logger');
 
 moment.suppressDeprecationWarnings = true; // We handle invalid date formats
 const dateFormat = 'YYYY-MM-DD';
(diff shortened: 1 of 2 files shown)


Dtphelan1:
Whoops this breaks more than it fixes – pushing up a more careful version of this change later 😄

jafeltra (approved):

PR #166, opened 2021-11-29
Removing severity and fixing actuality to 'actual'

dmendelowitz:
# Summary
This PR removes severity and fixes the value of actuality to 'actual' in the output of the CTCAdverseEventExtractor
## New behavior
Severity is no longer in the CSV template and is removed from the output of the client. Actuality is removed from the CSV template but will always have the fixed value of 'actual' in the output
## Code changes
- Severity and actuality removed from `CSVCTCAdverseEventExtractor.js`
- `CTCAdverseEventTemplate.js` updated to have severity removed and actuality fixed to 'actual'
- Related tests and test fixtures updated to reflect the changes
- All sample CSVs and `CSV_Templates.xlsx` updated to reflect changes
# Testing guidance
- Run the client with the CTC-AE Extractor and see that severity is no longer present in the output and that actuality always has the value of 'actual'
- Ensure that all tests still pass and that the updated tests correctly reflect the changes


jafeltra (approved):
Looks great to me! I couldn't think of any places you missed or any other tests to add!

PR #167, opened 2021-12-01
Adding event.coding.version and event.text to CTCAE Extractor

dmendelowitz:
# Summary
Adds an optional version field for the event coding and a required text field for the event of the CSVCTCAdverseEventExtractor
## New behavior
The CSV template for the CSVCTCAdverseEventExtractor now has a new required field of `adverseEventText` and a new optional field `adverseEventCodeVersion`
## Code changes
- `CTCAdverseEventTemplate.js` modified to add the new fields
- `CSVCTCAdverseEventExtractor.js` modified to have the new fields in the CSV template and check for new required field
- All example CSVs and `CSV_Templates.xlsx` changed to have new fields
- Relevant tests and test fixtures now test for the new fields
# Testing guidance
- Run the client with the CSVCTCAdverseEventExtractor and see that the event now has a coding version and text field
- Make sure the new fields work as described in the profile: https://build.fhir.org/ig/standardhealth/fsh-ae/StructureDefinition-ctc-adverse-event.html
- Ensure all tests still pass and that the test for the CTCAE properly test the new fields


dmendelowitz:
Turns out event.text is not required, so disregard anything said above about it being required

test/templates/ctcAdverseEvent.test.js
```diff
@@ -23,14 +25,16 @@ const VALID_DATA = {
 };
 
 const MINIMAL_DATA = {
-  // Only include 'id', 'subjectId', 'code', 'system', 'grade', and 'effectiveDateTime' fields which are required
+  // Only include 'id', 'subjectId', 'code', 'system', 'text', 'grade', and 'effectiveDateTime' fields which are required
```
julianxcarter, line 28: Super tiny, I think we can remove `text` from this comment right here now that it's no longer required.

test/templates/ctcAdverseEvent.test.js
```diff
@@ -43,13 +47,15 @@ const MINIMAL_DATA = {
 
 
 const INVALID_DATA = {
-  // Omitting 'subjectId', 'code', 'system', and 'effectiveDateTime' fields which are required
+  // Omitting 'subjectId', 'code', 'system', 'text', and 'effectiveDateTime' fields which are required
```
julianxcarter, line 50: Here as well.

julianxcarter (commented):
Looks great and works well! Just two minor suggestions.

julianxcarter (approved):

PR #169, opened 2021-12-09
Support CSVs with blank lines and empty values

Dtphelan1:
# Summary

Supports ingestion of CSV's with either totally blank rows of data or rows with nothing by empty-values; in both cases those rows will be ignored. 

## New behavior

Now the CSVFileModule can ingest files with rows that meet the above criteria.

## Code changes

Some new tests

# Testing guidance

Make sure the tool still works.

### DIFF
--- a/test/modules/CSVFileModule.test.js
+++ b/test/modules/CSVFileModule.test.js
@@ -4,7 +4,6 @@ const exampleResponse = require('./fixtures/csv-response.json');
 
 const INVALID_MRN = 'INVALID MRN';
 const csvFileModule = new CSVFileModule(path.join(__dirname, './fixtures/example-csv.csv'));
-const csvFileModuleWithBOMs = new CSVFileModule(path.join(__dirname, './fixtures/example-csv-bom.csv'));
 
 
 describe('CSVFileModule', () => {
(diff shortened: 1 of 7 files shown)


test/modules/CSVFileModule.test.js
```diff
@@ -4,7 +4,6 @@ const exampleResponse = require('./fixtures/csv-response.json');
 
 const INVALID_MRN = 'INVALID MRN';
 const csvFileModule = new CSVFileModule(path.join(__dirname, './fixtures/example-csv.csv'));
-const csvFileModuleWithBOMs = new CSVFileModule(path.join(__dirname, './fixtures/example-csv-bom.csv'));
```
dmendelowitz, line 7: I think we can delete `example-csv-bom.csv` now, right? It's not used anywhere else.

dmendelowitz (changes requested):
This looks great! Just had one small comment

jafeltra:
I'm not officially reviewing this PR, but when I was looking through the code yesterday and when Dylan M mentioned the bom nightmares, I remembered there are two other places we probably need these option as well. There's the `CSVURLModule` and the function in `appUtils` that uses it for the patient ids CSV. I think we want this in both of those places as well, right?

jafeltra:
> I wonder if it would be worth it to write a wrapper function that just calls `parse()` with the right settings and use that in all three places just so we never have to worry about this again?

Ah that's a good point. I feel like that could definitely be worthwhile.

test/modules/CSVFileModule.test.js
Dtphelan1: We are still using it, just encapsulated it within the test that uses it instead of leaving it up at the top

dmendelowitz: Sorry if I'm missing something obvious but I don't see the file used anywhere in the tests. The BOM test uses `example-csv-empty-values.csv`, is that intentional? I think that file has a BOM so it is an accurate test

Dtphelan1: Definitely not intentional. Definitely a typo. Fixing now.

dmendelowitz (approved):

PR #170, opened 2021-12-14
CSVExtractors support dataDirectory & fileName arguments

Dtphelan1:
# Summary

CSVExtractors now support ingesting a `dataDirectory` & `fileName` argument pair, with which the extractor constructs an absolute path to a data file. This technique for specifying the location of a CSV file on disk is now the preferred method, in contrast with our previous `filePath` argument technique. Documentation, example config files, and schemas have been updated to reflect this. 

## New behavior

All CSVExtractors extending from the BaseCSVExtractor should now support the ingestion of `fileName` and `dataDirectory` arguments. The BaseCSVExtractor now considers a new case in the initial extractor construction to handle the consumption of these arguments. 

## Code changes

- Updated the README to reflect this argument pair as our preferred approach;
- Updated example config to use this approach by default;
- Updated the BaseCSVExtractor to build that absolute file path when the two arguments are provided;
- Updated the BaseCSVExtractor to throw when the `dataDirectory` is not an absolute path;
- Updated all XCSVExtractors, where X is a ResourceType, to pass the `dataDirectory` and `fileName` arguments to their `super` constructors;
- Added `BaseCSVExtractor` to the extractors exported by `MEF/src/extractors/index.js`;
- Updated the config schema to reflect new properties, adding descriptions for human readability;
- Added tests to ensure that the BaseCSVExtractor handles its argument cases appropriately, throwing when expected.

# Testing guidance
- Ensure all tests pass;
- Ensure local extraction works once you've updated your Config appropriately; 
- Ensure that the schema updates are sufficient;
- Ensure that the approach is reasonable.
- Ensure that there aren't any missed documentation updates needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,8 @@ After exporting your CSV files to the `data` directory, kickstart the creation o
 2. `commonExtractorArgs.dataDirectory` should correspond to an absolute path to the dataDirectory containing all your exported CSV files;
 3. For each extractor, `fileName` should correspond to the file name this extractor should be reading from. Note: combining the `dataDirectory` above and `fileName` should resolve to a file on disk containing this corresponding extractor's data;
 
+**Note**: Previous versions of the MEF suggested using a `filePath` property for each extractor; while this property should still work without issue, the recommended approach is to use a common dataDirectory for all CSV files and to have each Extractor call out the name of the CSV file they need.
+
 For instructions on setting up an email notification trigger whenever an error is encountered in extraction, see the [Email Notification](#Email-Notification) section below.
 
 ### Configuration Deep Dive
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ To mask a property, provide an array of the properties to mask in the `construct
   "label": "patient",
   "type": "CSVPatientExtractor",
   "constructorArgs": {
-    "filePath": "./data/patient-information.csv"
+    "fileName": "patient-information.csv"
     "mask": ["address", "birthDate"]
   }
 }
@@ -157,7 +157,7 @@ Alternatively, providing a string with a value of `all` in the `constructorArgs`
   "label": "patient",
   "type": "CSVPatientExtractor",
   "constructorArgs": {
-    "filePath": "./data/patient-information.csv"
+    "fileName": "patient-information.csv"
     "mask": "all"
   }
 }
--- a/README.md
+++ b/README.md
@@ -92,8 +92,9 @@ Examples files for these extractor can be found in the [`test/sample-client-data
 
 After exporting your CSV files to the `data` directory, kickstart the creation of a configuration file by renaming the provided `csv.config.example.json` to `csv.config.json`. Then, ensure the following configuration parameters are properly set:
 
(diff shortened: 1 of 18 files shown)


src/extractors/BaseCSVExtractor.js
```diff
@@ -1,19 +1,36 @@
...
+  }) {
     super();
     this.unalterableColumns = unalterableColumns || [];
     this.csvSchema = csvSchema;
     if (url) {
+      logger.debug('Found url argument; creating a CSVURLModule with the provided url');
       this.url = url;
       this.csvModule = new CSVURLModule(this.url, this.unalterableColumns);
+    } else if (fileName && dataDirectory) {
+      if (!path.isAbsolute(dataDirectory)) throw new Error('dataDirectory is not an absolutePath, it needs to be.');
```
julianxcarter, line 19: Apparently, `path.isAbsolute` considers paths with a leading tilde (~) character to be a relative path. I guess this is technically correct, but it is something that I think we should allow. I'm not sure that writing our own function for checking that a path is absolute (while allowing the tilde character) is worth it, but if we decide to keep things as they are then we should change [the example dataDirectory property in csv.config.example.json](https://github.com/mcode/mcode-extraction-framework/blob/f1e2cdcfc426c32575e8fe7f2d0835b658e9c877/config/csv.config.example.json#L4) to use the `/Users/xxxxx/` paradigm instead of using a tilde.

julianxcarter (changes requested):
This looks great and I think the approach makes a lot of sense! I just have a few suggestions that I'd like your thoughts on :).

README.md
```diff
@@ -92,8 +92,9 @@ Examples files for these extractor can be found in the [`test/sample-client-data
 
 After exporting your CSV files to the `data` directory, kickstart the creation of a configuration file by renaming the provided `csv.config.example.json` to `csv.config.json`. Then, ensure the following configuration parameters are properly set:
 
-1. `patientIdCsvPath` should provide a file path to a CSV file containing MRN's for relevant patients;
-2. For each extractor, `filePath:` should provide a file path to a CSV file containing that corresponding extractor's data;
+1. `patientIdCsvPath` should correspond to an absolute file path to a CSV file containing MRN's for relevant patients;
+2. `commonExtractorArgs.dataDirectory` should correspond to an absolute path to the dataDirectory containing all your exported CSV files;
+3. For each extractor, `fileName` should correspond to the file name this extractor should be reading from. Note: combining the `dataDirectory` above and `fileName` should resolve to a file on disk containing this corresponding extractor's data;
```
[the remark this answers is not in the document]
Dtphelan1: 1) I think we're find for now omitting it, since the main reason we still support `filePath` is for backwards compatability. Alternatively, we could write a note saying that if your config uses a now-deprecated `filePath` property that we will still support it, without telling new people how to use that feature. Thoughts on that middle of the road approach? 
2) Great catch on the other examples. I will change those to use fileName.

src/extractors/BaseCSVExtractor.js
Dtphelan1: Great catch and great suggestion!

julianxcarter (approved):

PR #171, opened 2021-12-15
Participation Extension added

julianxcarter:
# Summary
Adds the [Participant extension](https://build.fhir.org/ig/standardhealth/fsh-ae/StructureDefinition-adverse-event-participant.html) to the CTC AE extractor.
## New behavior
The CTC AE extractor extractors two new values (`actor` and `functionCode`) that make up the two pieces of the Participant complex extension. If `functionCode` is included, `actor` must be included as well per the Ig.
## Code changes
1. Updated example values in `docs` and `sample-client-data`. New fields referenced in the `CSV_Templates` sheet as well.
2. Added the [Participant Function Value Set](https://build.fhir.org/ig/standardhealth/fsh-ae/ValueSet-adverse-event-participant-function-value-set.html) to the `valueSets` directory for use in fetching display values for the `functionCode` column.
3. Added support for the new columns to the CTC AE Template and Extractor.
4. Updated unit tests for the CTC AE Template and Extractor.
# Testing guidance
Ensure tests pass and extracted data looks good.


src/extractors/CSVCTCAdverseEventExtractor.js
```diff
@@ -34,10 +34,14 @@ function formatData(adverseEventData, patientId) {
...
       resolveddate: resolvedDate,
       seriousnessoutcome: seriousnessOutcome,
+      actor,
+      functioncode: functionCode,
     } = data;
 
     if (!(adverseEventCode && effectiveDate && grade)) {
       throw new Error('The adverse event is missing an expected attribute. Adverse event code, effective date, and grade are all required.');
+    } else if (functionCode && !actor) {
+      throw new Error('The adverse event is missing an expected attribute. Adverse event actor is a required element within the Participation extension');
```
dmendelowitz, line 44: This error message kind of makes it feel like actor is 100% required, not just conditionally required. I wonder if there's a better way to word it? Like would someone necessarily know what the participation extension is if they're just dumping a CSV into the client?

dmendelowitz (commented):
This works great! I just had one tiny comment about some wording

src/extractors/CSVCTCAdverseEventExtractor.js
julianxcarter: This is a really good point. I've changed this message to be a little more specific about the relationship between the `actor` and `functionCode` columns.

dmendelowitz (approved):

PR #172, opened 2021-12-17
Improving dataDirectory approach

Dtphelan1:
Better error messaging in the event that someone expects the DataDirectory to inform the patientIdCsv read operation


jafeltra (approved):
Looks good! Thanks for working through that with me!

PR #173, opened 2022-01-10
Fixed improper demorgans-law case logic for logger warning

Dtphelan1:
Logic is hard


dmendelowitz (commented):
The logic looks good but, is there any reason we aren't just checking they type of notification info to be boolean? Seems a little more fool proof

Dtphelan1:
> The logic looks good but, is there any reason we aren't just checking they type of notification info to be boolean? Seems a little more fool proof

This lateral thinking is why we do PR's. Yeah that would be way freaking easier. I'll swap now.

PR #178, opened 2022-03-02
Adverse Event context additions

julianxcarter:
# Summary
Adds two new utility functions, `getAdverseEventsFromContext` and `getAdverseEventEntriesFromContext`.
## New behavior
1. New additions to `contextUtils` make it possible to fetch AdverseEvent resources in the context bundle.
2. The `FHIRAdverseEventExtractor` will now only throw an error if a `study` is not included in its contructorArgs. Previously, an error would be thrown whenever there were no `ResearchStudy` resources found in context, even though it's not exactly necessary if a `study` is included.
## Code changes
1. Two utility functions added to `contextUtils.js`.
2. Any errors thrown by `getResearchStudiesFromContext` are now logged as debug messages in `FHIRAdverseEventExtractor.js`, which a custom error being thrown by that Extractor if the `study` contructorArg isn't included either.
# Testing guidance
Testing guidance will be included in the corresponding E-MEF PR.

### DIFF
--- a/src/extractors/FHIRAdverseEventExtractor.js
+++ b/src/extractors/FHIRAdverseEventExtractor.js
@@ -18,8 +18,12 @@ class FHIRAdverseEventExtractor extends BaseFHIRExtractor {
     try {
       allResearchStudyResources = getResearchStudiesFromContext(context);
     } catch (e) {
-      logger.error(e.message);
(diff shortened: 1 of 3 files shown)


dmendelowitz (approved):
Tested on the E-MEF, works as expected!

PR #183, opened 2022-07-08
CDS 2022 Value Set Updates

julianxcarter:
# Summary
Updates our Cancer Disease Status code lookups to include 2 new codes added in the [STU2 Condition Status Trend Value Set](https://hl7.org/fhir/us/mcode/ValueSet-mcode-condition-status-trend-vs.html).
## New behavior
The extractor now supports displayText mappings for the two new requested codes in the aforementioned VS. 
## Code changes
1. New codes added to the CDS lookup file. 
2. Code System check added to the CDSExtractor to ensure the proper system is used for the two additional codes. 
3. CDSExtractor test fixtures updated to test the new changes.   
# Testing guidance
1. Ensure unit tests pass. 
2. Run the extractor with the two new codes (`USCRS-352236` or `USCRS-352237`) in the `diseaseStatusCode` column of the `cancer-disease-status-information` csv and the `diseaseStatusText` column blank. The resulting JSON should include the new code, the corresponding display, and the corresponding system.

### DIFF
--- a/src/extractors/CSVCancerDiseaseStatusExtractor.js
+++ b/src/extractors/CSVCancerDiseaseStatusExtractor.js
@@ -36,7 +36,7 @@ class CSVCancerDiseaseStatusExtractor extends BaseCSVExtractor {
         status: observationStatus || 'final',
         value: {
           code: diseaseStatusCode,
-          system: 'http://snomed.info/sct',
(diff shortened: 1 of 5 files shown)


dmendelowitz (approved):
The new codes work as expected!

dmendelowitz:
@julianxcarter This looks like it's gonna merge into `main`, is there a reason for that or should we switch that to `develop`?

julianxcarter:
> @julianxcarter This looks like it's gonna merge into `main`, is there a reason for that or should we switch that to `develop`?

It should absolutely go into develop, thanks for catching that!

PR #185, opened 2022-12-15
Allowing CSVs with no rows + adding trimming to CSV parsing

dmendelowitz:
# Summary
This PR modifies the CSV validator function to allow CSVs with the proper header row, but no data to be considered valid. This PR also adds trimming to CSV parsing so excess whitespace in the CSV should be removed on parsing

## New behavior
When a CSV with no data is parsed, the parser returns an empty array, which meant the headers were not passed into the validator and so it would fail. This PR stores the headers separately to be passed into the validator so that even CSVs with no data will be properly validated.

Additionally, the `trim: true` option was added to the default options of the csv parse function, which should allow CSVs with extra whitespace to still work properly (`mrn         ,    familyName...` for example)

## Code changes
- New `getCSVHeader()` function added to `csvParsingUtils.js`, which returns an array of the headers from a CSV file
- `trim: true` added to `DEFAULT_OPTIONS` in `csvParsingUtils.js`
- `validateCSV()` in `csvValidator.js` modified to take headers as an input rather than infer them from `csvData`
- `CSVFileModule.js` and `CSVURLModule.js` modified to save the headers of CSVs they read and passthem as an input to `validateCSV()`
- Tests updated to reflect the above changes

# Testing guidance
- Ensure that extraction still works as expected with unaltered test data
- Add extra whitespace to CSVs and see that it is properly trimmed
- Remove data from CSVs and see that they pass as valid and extraction runs
- Modify CSV to have no data and improper headers and see that it is still properly caught as invalid

### DIFF
--- a/src/helpers/csvValidator.js
+++ b/src/helpers/csvValidator.js
@@ -27,10 +27,6 @@ function validateCSV(csvFileIdentifier, csvSchema, csvData, header) {
     });
   }
 
-  if (csvData.length === 0) {
-    return true;
-  }
-
   // Check values
   csvData.forEach((row, i) => {
     Object.entries(row).forEach(([key, value], j) => {
--- a/src/helpers/csvValidator.js
+++ b/src/helpers/csvValidator.js
@@ -3,11 +3,10 @@ const logger = require('./logger');
 
 // Validates csvData against the csvSchema
 // Uses the csvFileIdentifier in logs for readability
-function validateCSV(csvFileIdentifier, csvSchema, csvData, header) {
+function validateCSV(csvFileIdentifier, csvSchema, csvData, headers) {
   let isValid = true;
 
   // Check headers
-  const headers = header.map((h) => h.toLowerCase());
   const schemaDiff = _.difference(csvSchema.headers.map((h) => h.name.toLowerCase()), headers);
   const fileDiff = _.difference(headers, csvSchema.headers.map((h) => h.name.toLowerCase()));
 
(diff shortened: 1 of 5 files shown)


src/helpers/csvValidator.js
```diff
@@ -3,11 +3,11 @@ const logger = require('./logger');
 
 // Validates csvData against the csvSchema
 // Uses the csvFileIdentifier in logs for readability
-function validateCSV(csvFileIdentifier, csvSchema, csvData) {
+function validateCSV(csvFileIdentifier, csvSchema, csvData, header) {
   let isValid = true;
 
   // Check headers
-  const headers = Object.keys(csvData[0]).map((h) => h.toLowerCase());
+  const headers = header.map((h) => h.toLowerCase());
```
Dtphelan1, line 10: Any reason we don't include this normalization in the `getCSVHeader` function? I think I'd prefer any processing on the headers to occur in that function, so that if other utilities later need access to headers they're not all performing the normalization themselves.

Dtphelan1 (changes requested):
Other than my one comment, this looks pretty good. Can confirm that normal extraction works, the same data is extracted with whitespace on the CSV's, the extraction doesn't fail when data files are empty (except for headers), and invalid errors occur when headers are wrong with empty data.

src/helpers/csvValidator.js
dmendelowitz: That makes sense, should be addressed in the latest commit!

Dtphelan1 (approved):
looks good to me

PR #188, opened 2023-02-03
Modifying conditions to allow  multiple codes

dmendelowitz:
# Summary
This PR modifies the `CSVConditionExtractor` to accept multiple codes by separating them with a `|` (e.g. `1234|5678`). Adding extra codes will result in extra entries in the coding array of the resulting condition resource.

## New behavior
Multiple condition codes in the csv file, separated by a `|` should result in extra codes in the coding array of the output, all with the same system and display name. This should work for any number of additional codes, but adding `|` without a code (e.g. `123|`) should result in an error

## Code changes
- `CSVConditionExtractor.js` now splits codes into an array and maps them to all have the same system and display
- `ConditionTemplate.js` now takes an array of codes in `codingTemplate()` and maps those to an array in the output
- Tests updated to reflect changes

# Testing guidance
- Ensure that tests pass and extraction still works as expected
- Add extra codes to `condition-information.csv` and see that the output reflects the behavior described above (one code still works as expected, multiple codes results in extra codes in the output, improper use of `|` results in error)

### DIFF
--- a/src/extractors/CSVConditionExtractor.js
+++ b/src/extractors/CSVConditionExtractor.js
@@ -32,11 +32,11 @@ function formatData(conditionData, patientId) {
       subject: {
         id: patientId,
       },
-      code: {
-        code,
+      code: code.split('|').map((c) => ({
+        code: c,
         system: codeSystem,
         display: displayName,
-      },
+      })),
       category: category.split('|'),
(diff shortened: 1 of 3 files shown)


Dtphelan1 (approved):
Other than my one inline recommendation, this looks perfect. Works as expected on my local machine

PR #190, opened 2023-02-16
Trimming whitespace and handling empty values for additional condition codes

dmendelowitz:
# Summary

Trims whitespace and better handles empty values when using the `|` delimiter to include multiple condition codes

## New behavior

Whitespace is now trimmed off the split codes so inputs like `123 | 456`, etc, should no longer cause an error. Empty code values resulting from inputs like `|123` or `456|` should now print a warning to the output but still generate a condition resource with the one available condition (previously this would have caused an error and stopped the resource from being generated)

# Testing guidance
- Make sure extraction still works as expected and all tests still pass
- Test that the type of inputs described above work as described and don't cause any errors


Dtphelan1 (approved):
Works as expected locally, warning is readble. We'll deal with release hell tomorrow. 🎉

PR #193, opened 2023-05-05
Adding CSV schemas to remaining CSV extractors

dmendelowitz:
# Summary
Not every extractor had a CSV schema, so this PR adds schemas to those that did not. 

## New behavior
Missing CSV values would previously throw an error during extraction, now all our bases should be covered and we should be able to catch missing required fields in the validation stage and throw an error before extraction.

## Code changes
- `helper/schemas/csv.js` has new schemas for: `CSVStagingExtractor`, `CSVAdverseEventExtractor`, `CSVCTCAdverseEventExtractor`, `CSVCancerRelatedMedicationAdministrationExtractor`, `CSVCancerRelatedMedicationRequestExtractor`, `CSVObservationExtractor`, `CSVProcedureExtractor `
- The above listed extractors now add the proper schema during construction

# Testing guidance
- Make sure that the new schemas properly list all fields and mark the appropriate fields as required
- Make sure extraction still works as expected and missing fields are properly caught during the validation stage

### DIFF
--- a/src/extractors/CSVAdverseEventExtractor.js
+++ b/src/extractors/CSVAdverseEventExtractor.js
@@ -4,6 +4,7 @@ const { getEmptyBundle } = require('../helpers/fhirUtils');
 const { getPatientFromContext } = require('../helpers/contextUtils');
 const { formatDateTime } = require('../helpers/dateUtils');
 const logger = require('../helpers/logger');
+const { CSVAdverseEventSchema } = require('../helpers/schemas/csv');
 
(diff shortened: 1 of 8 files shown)


rdingwell (approved):
I see that the constraints are based on what we have in the extractors.  I can approve this now but we probably want to make sure the extractors are aligned to the spec.  For instance, in the medication request schema and extractor authoredOn is required, but it is not in the resource definition.

PR #197, opened 2023-09-26
Adding Period to ResearchSubjects in CSVClinicalTrialInformationExtractor

dmendelowitz:
# Summary

This PR adds the period field to the `ResearchSubject` generated by the `CSVClinicalTrialInformationExtractor` in the form of new optional `startDate` and `endDate` fields in the CSV template. Adding either or both fields to the CSV should result in a `period` in the output.

Additionally, this PR fixes `period` in the `CSVEncounterExtractor`. Accidentally made it so period would only show up if both start and end were present, but start and end are optional for periods, so you don't actually need both. This should work correctly now with just start or end

# Testing guidance
- Run the extractor and see that `ResearchSubject` resources in the output now have proper `period` fields
- Ensure that tests still pass and extraction still works as expected otherwise

### DIFF
--- a/src/templates/EncounterTemplate.js
+++ b/src/templates/EncounterTemplate.js
@@ -1,14 +1,5 @@
 const { ifAllArgsObj, ifSomeArgsObj } = require('../helpers/templateUtils');
-const { coding, reference } = require('./snippets');
-
-function periodTemplate({ startDate, endDate }) {
-  return {
-    period: {
-      ...(startDate && { start: startDate }),
-      ...(endDate && { end: endDate }),
-    },
-  };
-}
+const { coding, reference, periodTemplate } = require('./snippets');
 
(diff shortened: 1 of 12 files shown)


src/templates/EncounterTemplate.js
```diff
@@ -1,11 +1,11 @@
-const { ifAllArgsObj } = require('../helpers/templateUtils');
+const { ifAllArgsObj, ifSomeArgsObj } = require('../helpers/templateUtils');
 const { coding, reference } = require('./snippets');
 
 function periodTemplate({ startDate, endDate }) {
```
rdingwell, line 4: I think this should get moved to the templateUtils helper, it's used here and in researchsubject template.  It may be applicable elsewhere as well.

rdingwell (approved):

PR #199, opened 2023-10-26
Post-extraction mapper

dmendelowitz:
# Summary

This PR adds post-extraction filtering by date, which can be enabled in the config file. The user does this by specifying start and end dates and which resource types to filter and the post-extraction filter will only include resources within that range for those resource types.

Currently WIP is adding all the date formats for each resource type since they vary resource to resource, in the current state you can filter Observations and Conditions by date and I can easily expand it for other resource types.  In the current state, the date fields to be checked are hard-coded to be the style of date we produce with the CSV extractors, for example Observation could be any of the effectiveX dates, but I made it only check for effectiveDateTime since that is what we produce with the extractor. We could make a new task to create more robust date filtering that checks all possible date formats.

# Testing guidance
Add the following to the config file:
```
"postExtraction": {
    "dateFilter": {
      "startDate": "2019-01-01",
      "endDate": "2019-12-31",
      "resourceTypes": ["Condition", "Observation"]
    }
  },
```
Running the extractor with this filter should result in the first patient output having no Conditions and only one Observation left


dmendelowitz:
The latest commit changes how a mapper is included as a post-extraction process. Now, if the user includes a `mapper.js` file in their config folder that exports `resourceMapping` and `variables` then it will be included in the app and run as a new `AggregateMapper` post-extraction.
Here is an example mapper for testing purposes (should exclude patients named "Jane", which is the name of the patient in our first sample bundle output):
```
const fhirpath = require('fhirpath');
// eslint-disable-next-line camelcase
const fhirpath_r4_model = require('fhirpath/fhir-context/r4');

const resourceMapping = {
  filter: (r) => r.resource.resourceType === 'Patient',
  exclude: (r) => fhirpath.evaluate(r.resource, "name.given = 'Jane'", fhirpath_r4_model)[0],
  exec: (resource) => resource,
};

const variables = {};

module.exports = {
  resourceMapping,
  variables,
};
```

rdingwell (approved):

PR #202, opened 2023-11-07
Add search parameters

rdingwell:
# Summary
This PR adds the ability to configure search parameters for the FHIR extractors via construction arguments in the configuration file.  This allows us to potentially target specific resources and reduce the need for filtering data post extraction.  It also allows us to potentially use _include search filters to request additional information from a fhir server 
and reduce the number of requests we would need to make. 
## New behavior
Configuring search parameters in the construction arguments for FHIR extractors 
## Code changes
The FHIR extractors have been updated, the BaseFHIRExtractor class now accepts the new arguments and appends any search parameters to those returned with the patient id.   Removed all of the search params embedded in the extractors.  Updated the extractors to allow for the arguments and propagate to parent class.  Updated BaseFhirExtractorTest to test the new functionality and removed code testing  the embedded search parameters. 
# Testing guidance

### DIFF
--- a/src/extractors/FHIRPatientExtractor.js
+++ b/src/extractors/FHIRPatientExtractor.js
@@ -2,8 +2,8 @@ const { BaseFHIRExtractor } = require('./BaseFHIRExtractor');
 const { maskPatientData } = require('../helpers/patientUtils.js');
 
 class FHIRPatientExtractor extends BaseFHIRExtractor {
-  constructor({ baseFhirUrl, requestHeaders, version, mask = [] }) {
-    super({ baseFhirUrl, requestHeaders, version });
+  constructor({ baseFhirUrl, requestHeaders, version, mask = [], searchParameters }) {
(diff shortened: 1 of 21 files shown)


src/extractors/FHIRPatientExtractor.js
dmendelowitz, line 1: Is the `parametrizeArgsForFHIRModule` function in this file going to override the search parameters from the superclass version of the function?

rdingwell: That is the way it was coded before so I just left it as is.  Basically, there is no additional search functionality for Patient.

src/helpers/schemas/config.schema.json
```diff
@@ -41,6 +41,10 @@
           "title": "Request Headers",
           "type": "object"
         },
+        "queryParameters": {
```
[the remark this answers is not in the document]
rdingwell: I imagine they could go in either location.   There were other search parameters like 'context' that were in this location so that is where I put it.

